PR #5, opened 2013-08-12
New feature : filter issues with a minimum severity

sylvainlaurent:
In order to concentrate on most severe violations in eclipse, I'd like to hide less severe ones.
I added a dropdown in the preferences to allow the user to filter issues on a minimal severity, and modified the remote call to specify the severities to select.


henryju:
The feature still seems to be interesting but as we are trying to stop using SQ WS to retrieve issues the current implementation in this pull request can't be applied. Filtering should be done in SonarQube Issues view.

PR #6, opened 2013-08-13
Feature : default server loaded from preferences

axel3rd:
Hi, 

Proposition for this little feature to be able to load default SonarQube servers by -pluginCustomization eclipse process, using the proxy credentials if exists (in an organisation with centralised authentification system, the proxy and SonarQube authentification could be the same).

_Notes_ : 
* ScopedPreferenceStore can't iterate on a node but only get keys, so "new default system key" is required with index incrementation.
* Alone thread reference : [stackoverflow](http://stackoverflow.com/questions/17975557/sonarqube-eclipse-plugin-change-default-server-configuration)

_TODO (by me if idea accepted)_ : 
* Check why apache commons lang _StringUtils_ seems very long to load at first access
* Complete _org.sonar.ide.eclipse.core.tests_ for this feature (currently I develop with Eclipse Indigo and tycho integration gives me some compilation errors ... I need investigate with Kepler)

If you have suggestions or other strategy to realize this feature, I can change the orientation of my code.


henryju:
I have implemented this a bit differently (using regular Eclipse mechanism for default values). I didn't took the part about getting credentials from the proxy settings since in my opinion it is too heavily coupled with your organization internals). Anyway thanks for the PR.

PR #7, opened 2014-03-30
Let it be able to run in Eclipse 3.6

ks719:
Using new InstanceScope() instead of new InstanceScope() can let the
plugin run in Eclipse 3.6


henryju:
Support of Eclipse 3.6 is not our priority. And after a very quick look it seems InstanceScope.INSTANCE was introduced in Eclipse 3.4 so I don't see why your patch would solve the issue.

PR #8, opened 2014-05-12
fix: StackOverflowError when there is a circular dependency between projects

SchwingSK:
The bug appeared when at least 3 projects were involved: the first project depends on the second one which has a circular dependency towards the second one.
This prevented me from using the plugin on our biggest project (54 subprojects), whereas the Gradle sonar-runner plugin works flawlessly.

Attached code solves the bug and adds a test that reproduces the problem.
As far as my tests go, this doesn't seem to break anything.


henryju:
Manually applied (see http://jira.sonarsource.com/browse/SONARCLIPS-429). Thanks for your contribution.

PR #9, opened 2014-10-24
Enabled more control over source, test, libraries and binaries properties.

pwmedcraft:
Before these changes, the values for the following Sonar properties would be picked up from the project build path only:
sonar.sources
sonar.tests
sonar.binaries
sonar.libraries

If these properties were supplied in Project / Properties / SonarQube / Preview Analysis Properties, they were ignored and would not have been used in the scan. These changes stop such user-supplied properties being ignored. If they are supplied, their values will be appended to the values from the project build path.

However, these changes also enable the user to prevent the values from the project build path being included in the analysis. Project / Properties / SonarQube / Preview Analysis Properties now contains a checkbox for each of the properties. If the user leaves them checked, the respective settings from the project build path will be added to any user-supplied properties in the table above. If they are unchecked, only the user-supplied values in the property table will be used.

By default, the behaviour is at was before the changes.


TimJay:
This would be very helpful for us as well!

henryju:
Hi and sorry for this very late answer. I have chosen another option: user provided properties will always override properties automatically defined by configurators. This is more consistent with other bootstrappers like the Maven/Gradle plugins.
For example on a Java project, by default JavaProjectConfigurator will set sonar.sources/sonar.tests/sonar.libraries, ... but if user has defined the same properties in Project / Properties / SonarQube / Preview Analysis Properties then user props will override configurator props.

PR #10, opened 2014-12-12
Use slf4j osgi bundle for logging

bwnyasse:
To avoid breaking m2e with the following error [
java.lang.IncompatibleClassChangeError: Class
ch.qos.logback.classic.LoggerContext does not implement the requested
interface org.slf4j.ILoggerFactory] ,
this commit ensure that we used the slf4j ogsi bundle for logging.

see http://jira.codehaus.org/browse/SONARIDE-434


henryju:
I have removed any slf4j usage from SQ Eclipse plugin :)

PR #11, opened 2015-02-17
Allow user to override properties

daus-salar:
Allow users to override properties by adding extra properties at the end of analysis configuration.

I think this creates a better user experience because properties explicitly set by the user will be used instead of being silently ignored.

Furthermore:
* After a quick review I think this pull request addresses the same issue as mentioned in pull request #9 but is a bit smaller. 
* It will also enable users of the community c++ plugin to use the sonar-eclipse integration and will therefore solve the stackoverflow question: ["sonarqube analysis failed in eclipse throws error on local analysis due to the exception language “cpp” is not found"](http://stackoverflow.com/questions/21017860)

### DIFF
--- a/org.sonar.ide.eclipse.core.tests/src/test/java/org/sonar/ide/eclipse/core/internal/jobs/AnalyseProjectJobTest.java
+++ b/org.sonar.ide.eclipse.core.tests/src/test/java/org/sonar/ide/eclipse/core/internal/jobs/AnalyseProjectJobTest.java
@@ -140,13 +140,13 @@ public class AnalyseProjectJobTest extends SonarTestCase {
   }
 
   @Test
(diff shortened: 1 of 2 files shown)


henryju:
FYI I intend to do a major refactoring of SQ Eclipse to allow it to work with any language plugin. And this would indeed means to no more "force" some property values.

daus-salar:
This is nice to know, is there a scheduled release for this refactoring?

Is it an option to have this PR integrated to have a makeshift solution until your refactoring is done?

Regards Lars

henryju:
I won't commit on any schedule :). Concerning your PR I prefer to not apply it since this would require a release anyway and next release will definitely include this refactoring.

PR #12, opened 2015-03-19
SONARIDE-405 Added a new sonar configuration property 'sonar.tests.regexp'.

mebrookbank:
The new property allows the test sources regular expression to be overridden. If the property is not set
the default regular expression '.*test.*' is used.


henryju:
I have fixed SONARIDE-405 by introducing a new global configuration regexp. Anyway thanks for your contribution.

PR #13, opened 2015-07-16
SonarQube server association: Project configuration setup improvement

Hemantkumar-Chigadani:
SonaQube server association action and it's dialogue improvement when few of saved/persisted SonarQube servers are not reachable at given time. 

User should be allowed to associate a workspace project to a SonarQube server if at-least one of SonarQube servers is reachable, with additional warning description shown on the association dialogue about non reachable servers details.

Fix to a Issue: You can not associated a SonarQube server to a workspace project if any one of saved/persisted SonarQube servers is not reachable.


henryju:
Could you please remove all 'final' on local variables. This is not in our coding practice and it makes your P/R hard to read. Thanks!

Hemantkumar-Chigadani:
Thanks for the comments.Removed all 'final' on local variables.

I think, Luna Eclipse does not job well when multiple Java formatters exist and applied to different projects in the same workspace.

Hemantkumar-Chigadani:
Please refer to new pull request #19.

PR #15, opened 2015-07-16
Lazy caching of Sonar servers version.

Hemantkumar-Chigadani:
Lazying caching of SonarQube server version improves performance if the client code is interested in servers details except version. Example User wants to edit the servers list in Eclipse preference when SonarQube server is not reachable.
Version caching and it's parsing has been encapsulated to a new Java file 'Version'.

Please note this improves the performance considerably when more than one SonarQube servers are  configured by user to analysis code. This helps a lot in auto-sonar-builder[Future concept] too.

### DIFF
--- a/org.sonar.ide.eclipse.core.tests/src/test/java/org/sonar/ide/eclipse/core/internal/jobs/AnalyzeProjectJobTest.java
+++ b/org.sonar.ide.eclipse.core.tests/src/test/java/org/sonar/ide/eclipse/core/internal/jobs/AnalyzeProjectJobTest.java
@@ -19,11 +19,15 @@
  */
 package org.sonar.ide.eclipse.core.internal.jobs;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.fail;
+
 import java.io.File;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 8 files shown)


henryju:
Server version and status is cached in master. It is "refreshed" when accessing server settings page but I think that's ok to wait a little here. See SONARCLIPS-407

Hemantkumar-Chigadani:
Thanks for the comments. I agree with description given in the SONARCLIPS-407. But I think, I was trying to do the SonarServer reachability improvement, nothing to do with the issue reported.I was coming to ISonarServer and it's implementation SonarServer from completely different angle when I was coding for this branch.
With old implementation, SonarServer version caching mechanism gave us a way of checking server reachability status dynamically whenever a user requests for the code analysis as soon as any associated SonarServer becomes reachable without restarting Eclipse.
With present implementation disabled()/setDisabled() on SonarServer/ISonarServer, Servers reachability status update is limited only once during user first code analysis request. User needs to restart Eclipse or reload preference for SonarServers list to check and update the SonarServer reachability status.
I hope you agree,old implementation gave us more dynamic approach towards server reachability.

I would like to embed dynamical/live server reachability checks within present implementation of disabled() instead of using a flag if you could give me few more days. Let me know what you think.

PR #17, opened 2015-07-18
SONARCLIPS-403 user improvment

gedeffe:
Hello,
Looking to the bug SONARCLIPS-403, I found it relevant with following use case :
when working offline, the sonar console pops every time I open a java file. And because, this view is located in same area than problems view, I had to switch back manually to problems view.

To avoid that, an easy workaround is to disable notification from sonar console. Yet, I have to switch explicitly when I want issues update from server.

A little improvement would be to log into the console only on explicit user's action and use eclipse error log for others. So, the sonar console would pops only when I need it :-)

Please have a look to these few lines as it would be easy to integrate into this task.
I hope it could help.

Thanks,
Gabriel

PS: sorry for the noise with 2 pull request to update the fork. On the other side, it should ease the merge.


henryju:
Thanks for your contribution and sorry for the late answer. We focused all our efforts on new SonarLint plugin so I will not apply this P/R. Please test SonarLint (to be annouced in the coming days) and feel free to come back if you still think the console is annoying (I tried to improve that compared to old plugin).

PR #19, opened 2015-07-20
Project configuration dialog with more details about non-reachable servers for the user.

Hemantkumar-Chigadani:
Updated pull request for  #13.


henryju:
We have started a new plugin called SonarLint. For now there is no server association but this feature will be back soon. At that time I will be careful to not fail badly when one server is not reachable (using good ideas from your P/R). Thanks for this contribution.

PR #40, opened 2016-10-26
New issue tracking, step 1

janos-ss:
Note: the PR target is not `master`, but `feature/issue-tracking` (a copy of `master`).

What should work:

- Issue tracking in standalone or connected mode, except matching by checksum (text range hash and line hash are broken at the moment)
- Hide resolved issues

Regressions (prevent merging to `master`):

- Trigger analysis only on file open and on explicit command (currently Save triggers it too)
- Fix matching by text-range-hash (see `AnalyzeProjectJob.computeChecksum` for a hint)
- Fix marker severity display in issues view
- Fix marker display in editor view
- Make `IssueTrackerCache` persistent, initialized at startup
- Fix broken tests (for example, https://travis-ci.org/SonarSource/sonarlint-eclipse/jobs/170761446)

Other TODO:

- Improve the initialization of the per-project `IssueTracker` instances: should happen on project import / open event
- Implement matching by line hash using MD5 as on the server (is the md5 implementation osgi-ready?)
- Confirm correct resolution of module key and file path in nested maven project
- Download and store server issues on binding
- Unit tests
- Other smelly stuff (for example, `Logger` and `Console` in `org.sonarlint.eclipse.core.internal.tracking` package)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AnalyzeProjectJob.java
```diff
@@ -342,71 +299,55 @@ private void updateMarkers(Map<IResource, List<Issue>> issuesPerResource, Analys
...
-      } else {
-        creationTimeStampForNewIssues = now;
-      }
       List<Issue> rawIssues = resourceEntry.getValue();
       try {
-        if (r instanceof IFile) {
-          issueTrackingOnFile(iTextFileBufferManager, r, previousMarkers, rawIssues, creationTimeStampForNewIssues);
+        if (resource instanceof IFile) {
+          IProject project = resource.getProject();
+          String moduleKey = project.getName();
```
henryju: moduleKey = Eclipse project name ??

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AnalyzeProjectJob.java
```diff
@@ -342,71 +299,55 @@ private void updateMarkers(Map<IResource, List<Issue>> issuesPerResource, Analys
...
+          SonarLintCorePlugin.getDefault().getModulePathManager().setModulePath(moduleKey, project.getLocation().toString());
+
+          String relativePath = resource.getProjectRelativePath().toString();
+          trackLocalIssues(moduleKey, resource, relativePath, rawIssues);
+          trackServerIssues(moduleKey, resource, relativePath);
         } else {
-          issueTracking(r, previousMarkers, rawIssues, null, creationTimeStampForNewIssues);
+          // TODO delete if this never happens, or else handle it better
```
henryju: I think this part was to support issues reported at project level (for example rule about missing package-info.java)

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AnalyzeProjectJob.java
janos-ss: It's used for local, per-module storage of tracking data. As long as Eclipse project names are unique per workspace, it should be fine. Renaming projects will need some extra work.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AnalyzeProjectJob.java
janos-ss: Thanks, I will try to do something better than just crash the application

dbmeneses:
Looking good so far

PR #58, opened 2016-11-09
SLE-110 Restore the handling of file level issues (= no text range)

janos-ss:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/markers/MarkerUtils.java
```diff
@@ -75,7 +75,7 @@ public static void updateAllSonarMarkerSeverity() throws CoreException {
 
   @CheckForNull
   public static FlatTextRange getFlatTextRange(final IDocument document, TextRange textRange) {
```
henryju: add @Nullable on testRange then?

PR #70, opened 2016-11-28
Run Analysis on Entire Project Enabled

amitbhoraniya:
If user want to analysisi  entire project from eclipse then it is currently not enable, user need to open each and every file to see weather  there is any issues or not, this may useful to user to scan entire project and its really useful feature.
Discussion url for that issue is: https://groups.google.com/forum/#!msg/sonarlint/N2Qaxh58XI0/AWzg81ljAgAJ;context-place=topic/sonarlint/izf46IVhvSo

### DIFF
--- a/org.sonarlint.eclipse.ui/plugin.xml
+++ b/org.sonarlint.eclipse.ui/plugin.xml
@@ -387,15 +387,24 @@
 	      <visibleWhen>
 	         <with variable="activeMenuSelection">
 	            <iterate ifEmpty="false">
-	              <and>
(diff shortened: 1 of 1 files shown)


henryju:
Hi, thanks for your suggestion. We have plans to allow analyzing all files but that will probably be something similar to the "analyze all changed files" new feature. You can follow our progress on this ticket: https://jira.sonarsource.com/browse/MMF-639
I'm closing this P/R since this is not the way we will implement it.

PR #73, opened 2016-12-05
SLE-128 Wrap store reader in try-with-resources

janos-ss:


janos-ss:
The other (more important) part of the fix of SLE-128 will be changing the dependency version to the latest core.

janos-ss:
This branch is pointless now that 2.4 is out. I will create a new PR with the complete fix.

PR #119, opened 2017-09-18
Test travis utils with local mirror

janos-ss:


janos-ss:
No rush with this, we can wait for the next release of `travis-utils` to use proper version numbers.

PR #120, opened 2017-10-26
Feature/analyze after build

henryju:


its/target-platform-e43/target-platform-its-e43.target
```diff
@@ -17,7 +17,7 @@
 </location>
 <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
 <unit id="org.python.pydev.feature.feature.group" version="0.0.0"/>
-<repository location="http://www.pydev.org/update_sites/5.2.0"/>
+<repository location="https://dl.bintray.com/fabioz/pydev/5.2.0"/>
```
janos-ss: Why is this better?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintPostBuildListener.java
```diff
@@ -48,11 +48,11 @@
 import org.sonarlint.eclipse.ui.internal.markers.ShowIssueFlowsMarkerResolver;
 import org.sonarlint.eclipse.ui.internal.util.PlatformUtils;
 
-public class SonarLintChangeListener implements IResourceChangeListener {
+public class SonarLintPostBuildListener implements IResourceChangeListener {
 
   @Override
   public void resourceChanged(IResourceChangeEvent event) {
-    if (event.getType() == IResourceChangeEvent.POST_CHANGE) {
+    if (event.getType() == IResourceChangeEvent.POST_BUILD) {
```
janos-ss: Will this work for languages that are not built, such as JavaScript?

its/target-platform-e43/target-platform-its-e43.target
henryju: It was failing on cix and I don't understand why. Seems related to the redirect + composite update site.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintPostBuildListener.java
henryju: I did a test with JavaScript, and QA is passing. Also, documentation says the POST_BUILD event should be triggered at the time build would have been done, independently of the fact a build(er) actually run.

PR #128, opened 2017-11-15
SLE-218 Restore support of Visual Cobol

henryju:


org.sonarlint.eclipse.jdt/src/org/sonarlint/eclipse/jdt/internal/JdtUtils.java
```diff
@@ -60,11 +63,21 @@ static boolean hasJavaNature(IProject project) {
...
     IJavaElement javaElt = JavaCore.create(file);
-    return javaElt == null || (hasJavaNature && isStructureKnown(javaElt) && javaProject.isOnClasspath(javaElt));
+    if (javaElt == null) {
+      // Not a Java file, don't exclude it
+      return false;
+    }
+    if (!javaElt.exists()) {
+      // SLE-218 Visual Cobol with JVM Development make JDT think .cbl files are Java files. But still we want to analyze them, so only
+      // exclude
+      // files having the original java source content type.
```
janos-ss: Cosmetic, but this comment would be easier to read if layed out like this:

    // SLE-218 Visual Cobol with JVM Development make JDT think .cbl files are Java files.
    // But still we want to analyze them, so only exclude files having the original java source content type.

PR #134, opened 2018-02-13
Update slcore

dbmeneses:


org.sonarlint.eclipse.ui/plugin.xml
```diff
@@ -619,6 +659,21 @@
...
     </handler>
+    <handler
+          class="org.sonarlint.eclipse.ui.internal.command.ExcludeCommand"
+          commandId="org.sonarlint.eclipse.ui.internal.command.ExcludeCommand">
+           <enabledWhen>
+	        <with variable="activeMenuSelection">
+	          <iterate ifEmpty="false">
+	            <or>
+                 <adapt type="org.sonarlint.eclipse.core.resource.ISonarLintFile">
+                   <test property="org.sonarlint.eclipse.core.notExcluded"/>
```
henryju: I wonder if it is possible to revert the property, and use a negation. Like:
```
<not>
    <test property="org.sonarlint.eclipse.core.excluded"/>
</not>
```

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintPartListener.java
```diff
@@ -64,7 +68,7 @@ private static void scheduleUpdate(IEditorPart editorPart, ISonarLintFile sonarL
 
   private static void scheduleUpdate(FileWithDocument fileWithDoc) {
     ISonarLintFile file = fileWithDoc.getFile();
-    if (!file.getProject().isAutoEnabled()) {
+    if (!file.getProject().isAutoEnabled() || !FileExclusionsUtils.shouldAnalyze(file.getProject(), file.getResource())) {
```
henryju: I'm not comfortable with the use of file.getProject() and file.getResource(). Especially because you are later using their path to decide on the exclusion. Won't it be possible to use the ISonarLintFile::getProjectRelativePath() since this is the "logical" path.

org.sonarlint.eclipse.ui/plugin.xml
dbmeneses: It is, I will try it

org.sonarlint.eclipse.ui/plugin.xml
```diff
@@ -664,7 +663,7 @@
           commandId="org.sonarlint.eclipse.ui.internal.command.ExcludeCommand">
            <enabledWhen>
 	        <with variable="activeMenuSelection">
-	          <iterate ifEmpty="false">
+	          <iterate ifEmpty="false" operator="or">
 	            <or>
```
henryju: What is the point of this "or". Shouldn't be a and?

dbmeneses: In fact I think it's a no-op in this case, since there is only on adapt being done. The exclusion test is part of the adapt.

henryju: Arf, you're right, I was confused by the indentation ;)

PR #166, opened 2018-06-06
Apply feedback

janos-ss:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/RulesConfigurationPage.java
```diff
@@ -47,7 +47,7 @@ public RulesConfigurationPage() {
 
   @Override
   public void init(IWorkbench workbench) {
-    setDescription("Configure active rules in standalone mode.");
+    setDescription("Configure active rules in standalone mode. When a project is connected to a SonarQube server, configuration from the server applies.");
```
henryju: What about SonarCloud?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/RulesConfigurationPart.java
```diff
@@ -201,7 +201,7 @@ private void createTreeViewer(Composite parent) {
   private void createRuleDescriptionPart(Composite parent) {
     try {
       Browser browser = new Browser(parent, SWT.FILL);
-      browser.setText("(No rule selected.)");
+      browser.setText("<small><em>(No rule selected)</em></small>");
```
henryju: Is it not supposed to be "no rule**s**" ?

janos-ss: In terms of grammar, I think both singular and plural would be correct, but mean slightly different things. Plural would imply that a multi-selection is possible, but in fact it won't make a difference for rule display, since only one rule will be displayed at a time. I think singular will be less confusing.

henryju: OK, I trust you :)

PR #174, opened 2018-10-09
SLE-242 Improve issue tracking when having very repetitive issues

henryju:
* first match using the server issue key
* in case of no match with a server issue, keep the local leak


henryju:
@janos-ss since no tests are failing after moving it first, then I assume it is ok :)

More seriously, one problem I think we will face is that once an issue has been wrongly matched, the algorithm will never "recover" since we will always prefer to match on the serverIssueKey.

I discussed with @christophelevis  this concern, but I think I will wait for a real life use case before trying to improve things.

PR #176, opened 2019-01-03
Feature - Toggle editor change and editor open trigger

MarthinusKruger:
## Description of Problem:

Currently we have a bunch of class files in our project that exceed even 5000+ lines of code (legacy code) that takes about 20+ seconds to finish an analysis (when triggered).  In terms of making changes to these files (causing save action trigger) and opening various class files during an issue investigation the analysis gets triggered often causing the IDE the to freeze frequently causing unnecessary time to be lost.

## Console Log of Problem:

```
Trigger: EDITOR_OPEN
Clear markers on 0 excluded files
SonarLint processing file /fcbackend/src/main/java/com/fc/api/BaseApi.java...
Connected mode (using configuration of '###' in server '#########')
Starting analysis with configuration:
...
Forced to init a physical file
...
Found 382 issue(s)
Done in 27856 ms
```

```
Starting SonarLint for Eclipse 4.0.0.201810170711
Trigger: STARTUP
Clear markers on 0 excluded files
SonarLint processing file /fcbackend/src/main/java/com/fc/api/WorkflowApi.java...
Connected mode (using configuration of '###' in server '#########')
Starting analysis with configuration:
...
Forced to init a physical file
...
Check for updates from server '#########'
Found 544 issue(s)
0 entries removed from the store
Done in 34767 ms
```

## Proposed Solution:

The proposed solution would be to provide the user/developer the ability to "toggle" the `EDITOR_CHANGE `and `EDITOR_OPEN` analysis triggers.  Having this option on a per project level (part of the project's configuration) satisfies both use cases presented above.  This would mean a persisted mechanism to easily flag the state of the trigger to either `true` or `false` depending on the context of its use.  

## Changes Implemented:

1. Added two new command functions `Toggle XXX Analysis` option on project right-click view.
2. Add two new flags to the `SonarLintProjectConfiguration` class.
3. Added the mechanism to get the current state of the trigger type, inverse, and apply to the project's preferences (the toggle mechanism).
4. Added the necessary checks when an analysis is triggered that the appropriate flag is also checked to determine if the analysis should be scheduled or not.
5. Appropriate logging for both the toggling of the trigger state as well as a analysis being triggered but the state being false to provide the necessary indication of such actions.

## Further Comments:

I have not yet added the icons for the two new commands created.  I am not sure how the other icons were generated or where they we downloaded from so that is still outstanding; this will be added if the proposed solution is acceptable.

Let me know if there is any concern with my description above or my implementation displayed below in terms of how I implemented it.  I have no problem discussing this at length and any questions are welcome.  This is my first potential contribution to the sonarlint-eclipse project so if there is anything in terms of standard ways of implementation I missed please let me know.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/SonarLintProjectConfiguration.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/SonarLintProjectConfiguration.java
@@ -23,9 +23,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
-
 import javax.annotation.Nullable;
-
 import org.sonarsource.sonarlint.core.client.api.connected.ProjectBinding;
 
 public class SonarLintProjectConfiguration {
@@ -92,7 +90,7 @@ public class SonarLintProjectConfiguration {
 
   /**
    * 
-   * @param analysisEditorTriggerEnabled
+   * @param triggerEditorOpenEnabled
    */
   public void setTriggerEditorOpenEnabled(boolean triggerEditorOpenEnabled) {
     this.triggerEditorOpenEnabled = triggerEditorOpenEnabled;
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/SonarLintProjectConfiguration.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/SonarLintProjectConfiguration.java
@@ -23,7 +23,9 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+
 import javax.annotation.Nullable;
+
 import org.sonarsource.sonarlint.core.client.api.connected.ProjectBinding;
 
 public class SonarLintProjectConfiguration {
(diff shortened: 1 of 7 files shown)


henryju:
Hi @MarthinusKruger,

Thanks for your interest in SonarLint. Before discussing your proposal, I would like to clarify two things.

1) Why not try to fix the performance issue, before cutting some features?
In the past, we managed to fix huge performance issue in SonarLint. The difficulty is always the same: have a good reproducer so that we could investigate.
First, it would be interesting to know more precisely in which component the time is spent. Would you mind sharing a bit more of the analysis logs? I'm especially interested by logs like:
```
Execute Sensor: JavaSquidSensor
Configured Java source version (sonar.java.source): 8
JavaClasspath initialization
JavaClasspath initialization (done) | time=8ms
JavaTestClasspath initialization
JavaTestClasspath initialization (done) | time=9ms
[...]
Java Main Files AST scan
x source files to be analyzed
[...]
Java Main Files AST scan (done) | time=55ms
Java Test Files AST scan
1/1 source files have been analyzed
0 source files to be analyzed
0/0 source files have been analyzed
Java Test Files AST scan (done) | time=1ms
```
Would be also interesting to know:
* your JDK version
* SonarJava version (since you are using connected mode)
* the size of your classpath (number of entries in sonar.java.libraries)


2) We already have an option to disable on-the-fly analysis (Run SonarLint automatically in project settings). I'm a bit skeptical to introduce more configuration to choose between the EDITOR_CHANGE and EDITOR_OPEN triggers. If the analysis is so slow that it is annoying, in my opinion you don't want it to run automatically at all.

henryju:
I'm closing this due to low activity. Feel free to comment to answer my previous questions.

PR #184, opened 2019-06-05
Reduced scope of workspace operations

jhonnen:
Only run actual resource change (i.e. marker updates) within a workspace
operation to avoid blocking the autobuild during long-running analyses.

https://community.sonarsource.com/t/sonarlint-analysis-blocks-autobuild/10483

I don't think the changes can be unit tested (sensibly).

### DIFF
--- a/its/org.sonarlint.eclipse.its/src/org/sonarlint/eclipse/its/utils/JobHelpers.java
+++ b/its/org.sonarlint.eclipse.its/src/org/sonarlint/eclipse/its/utils/JobHelpers.java
@@ -75,6 +75,7 @@ public final class JobHelpers {
           jobs[i].join();
         }
       }
+      jobManager.join("org.sonarlint.eclipse.projectJob", null);
       workspace.run(new IWorkspaceRunnable() {
         @Override
         public void run(IProgressMonitor monitor) {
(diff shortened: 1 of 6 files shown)


henryju:
This is breaking our integration tests, that are waiting for background job to complete before checking if markers are correct. We would have to find a better strategy. Do you have any idea?
The current strategy is:
https://github.com/SonarSource/sonarlint-eclipse/blob/fb95c5de189d1374e8492f745c1b92dafa610792/its/org.sonarlint.eclipse.its/src/org/sonarlint/eclipse/its/utils/JobHelpers.java#L38

jhonnen:
It's probably enough to update the job matching, since the analyse jobs aren't WorkspaceJobs anymore.

https://github.com/SonarSource/sonarlint-eclipse/blob/fb95c5de189d1374e8492f745c1b92dafa610792/its/org.sonarlint.eclipse.its/src/org/sonarlint/eclipse/its/utils/JobHelpers.java#L74

henryju:
Thanks, ticket created, I will do some more tests and merge if everything is fine: https://jira.sonarsource.com/browse/SLE-307

henryju:
Merged, thanks a lot!

PR #187, opened 2019-08-29
Hardening

henryju:


henryju:
@dbmeneses I will assign tickets for testing later. I think it is simpler to do a single pass once all tickets will have been handled.

PR #188, opened 2019-09-02
Hardening week 2

henryju:


henryju:
> The icon definitely looks nicer in the "Show View" menu and in the tab title. Can we do something about the context menu item here?

This one is even lower resolution, so I don't know what I can do. I already took the same question mark than an official Eclipse icon: https://github.com/eclipse/eclipse.platform.ua/blob/1c74f30ab93c0c3f14ece7fd43f1c481ffa2ed91/org.eclipse.help.ui/icons/etool16/helpsearch_co.gif

PR #196, opened 2019-10-30
SLE-320 Hook up to e4 dark theme for issue location highlighting

jblievremont:


org.sonarlint.eclipse.ui/plugin.xml
```diff
@@ -929,5 +929,14 @@
            class="org.sonarlint.eclipse.ui.internal.preferences.SonarLintPreferencesInitializer">
      </initializer>
   </extension>
-	
+
+  <extension
+        point="org.eclipse.e4.ui.css.swt.theme">
```
henryju: Any constraint on the minimal version of Eclipse?

jblievremont: AFAICT, the E4 CSS-based theming capabilities are GA since 4.2 (Juno); I traced back the earliest commits on this extension point to 2010, so I am rather confident about support in our targeted versions.

henryju: Great! I wonder if we could also use the same technic to improve rendering of other colors :)

PR #200, opened 2019-11-13
SLE-321 Allow use of secondary locations for issues of the report mode

jblievremont:


jblievremont:
Fixed as part of #265

PR #204, opened 2019-11-22
Restore coverage

jblievremont:


org.sonarlint.eclipse.site/pom.xml
```diff
@@ -27,4 +27,37 @@
...
+            <artifactId>jacoco-maven-plugin</artifactId>
+            <executions>
+              <execution>
+                <id>report-aggregate</id>
+                <phase>prepare-package</phase>
+                <goals>
+                  <goal>report-aggregate</goal>
+                </goals>
+                <configuration>
+                  <title>SonarLint Eclipse</title>
```
henryju: Not needed for us

org.sonarlint.eclipse.site/pom.xml
```diff
@@ -27,4 +27,37 @@
     </plugins>
   </build>
 
+  <profiles>
+    <profile>
+      <id>coverage</id>
+      <dependencies>
+        <dependency>
```
henryju: I would move this profile to the tests module, in order to avoid having to add this dependency.

PR #217, opened 2020-02-13
SLE-335 Apply test file regexp to all files not already considered as tests

jblievremont:


jblievremont:
Yup, I tried to write a medium test but this is more probably worth an IT.

its/org.sonarlint.eclipse.its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
```diff
@@ -139,6 +146,14 @@ public void shouldAnalyseJavaJunit() throws Exception {
       tuple("/java-junit/src/hello/Hello.java", 11, "Replace this use of System.out or System.err by a logger."),
       tuple("/java-junit/src/hello/Hello.java", 15, "Remove this unnecessary cast to \"int\".")); // Test that sonar.java.libraries is set
 
+    new JavaPackageExplorerBot(bot)
+      .expandAndDoubleClick("java-junit", "src", "hello", "HelloTestUtil.java");
+    JobHelpers.waitForJobsToComplete(bot);
+  
+    List<IMarker> testUtilMarkers = Arrays.asList(project.findMember("src/hello/HelloTestUtil.java").findMarkers(MARKER_ON_THE_FLY_ID, true, IResource.DEPTH_ONE));
+    // File is flagged as test by regexp, no issue raised
```
henryju: I would find it safer if you could make some test specific issues appearing on this file (like missing assertion), just to really confirm it is analyzed as test (and not simply excluded).

PR #228, opened 2020-05-15
SLE-353 Revert changes of SLE-344

henryju:


damien-urruty-sonarsource:
:+1:

damien-urruty-sonarsource:
As [SLE-344](https://jira.sonarsource.com/browse/SLE-344) was supposed to speed things up, do we have another way to do so or are locked with Eclipse's own performance ?

henryju:
The (only) guy reporting the issue investigated the problem down to eGit (Eclipse integration with Git). There is maybe an issue on this side, but for now it is safer to revert to the previous state.

PR #233, opened 2020-05-28
SLE-352 Improve SonarCloud usage report in telemetry

damien-urruty-sonarsource:
We report SonarCloud usage if at least one project is bound and opened.

Previously we could report SonarCloud usage if a server connection to SC was created but no project was bound.

### DIFF
--- a/org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/telemetry/SonarLintTelemetryTest.java
+++ b/org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/telemetry/SonarLintTelemetryTest.java
@@ -63,19 +63,12 @@ public class SonarLintTelemetryTest extends SonarTestCase {
 
   @Before
   public void start() throws Exception {
-    // Clear property that is set in the surefire arguments (see pom.xml)
-    System.clearProperty(SonarLintTelemetry.DISABLE_PROPERTY_KEY);
     this.telemetry = createTelemetry();
 
     project.open(MONITOR);
     ROOT.node(PREF_SERVERS).removeNode();
     SonarLintCorePlugin.getServersManager().stop();
     SonarLintCorePlugin.getServersManager().init();
-  }
-
-  @After
-  public void after() {
-    System.clearProperty(SonarLintTelemetry.DISABLE_PROPERTY_KEY);
   }
 
   private SonarLintTelemetry createTelemetry() {
@@ -91,11 +84,8 @@ public class SonarLintTelemetryTest extends SonarTestCase {
   }
 
   @Test
-  public void disable_property_should_disable_telemetry() throws Exception {
-    assertThat(createTelemetry().enabled()).isTrue();
-
-    System.setProperty(SonarLintTelemetry.DISABLE_PROPERTY_KEY, "true");
-    assertThat(createTelemetry().enabled()).isFalse();
+  public void telemetry_should_be_deactivated_for_tests() {
+    assertThat(SonarLintTelemetry.shouldBeActivated()).isFalse();
   }
 
   @Test
--- a/org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/telemetry/SonarLintTelemetryTest.java
+++ b/org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/telemetry/SonarLintTelemetryTest.java
@@ -20,9 +20,22 @@
 package org.sonarlint.eclipse.core.internal.telemetry;
 
 import java.nio.file.Path;
+
+import org.eclipse.core.resources.IProject;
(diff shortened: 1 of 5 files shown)


org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/telemetry/SonarLintTelemetryTest.java
```diff
@@ -31,16 +44,34 @@
...
     System.clearProperty(SonarLintTelemetry.DISABLE_PROPERTY_KEY);
     this.telemetry = createTelemetry();
+
+    project.open(MONITOR);
+    ROOT.node(PREF_SERVERS).removeNode();
+    SonarLintCorePlugin.getServersManager().stop();
+    SonarLintCorePlugin.getServersManager().init();
   }
 
   @After
```
henryju, line 77: I know this is not part of your changes, but I wonder if there is not an issue here. Shouldn't we restore the 'disable' property to 'true' to ensure next test suites do not send ping to the real telemetry?

org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/telemetry/SonarLintTelemetryTest.java
```diff
@@ -157,4 +188,67 @@ public void analysisDoneOnSingleFile_should_not_trigger_analysisDoneOnSingleFile
...
+  @Test
+  public void should_not_report_sonar_cloud_usage_when_project_is_bound_to_localhost() {
+    addServer("localhost", "http://localhost:9000");
+    bindImportedProjectToServer("localhost");
+
+    assertThat(SonarLintTelemetry.isAnyOpenProjectBoundToSonarCloud()).isFalse();
+  }
+
+  @Test
+  public void should_report_sonar_cloud_usage_when_project_is_bound_to_sonar_cloud() {
```
henryju, line 208: To me one test is missing, the one at the source of the ticket: `should_not_report_sonar_cloud_usage_when_sonarcloud_connection_but_project_is_not_bound`

damien-urruty-sonarsource: Interesting ! It was deliberate. But in the end it might be better to have it

org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/telemetry/SonarLintTelemetryTest.java
damienurruty: I moved the check for the system property up in the call hierarchy, so that in the tests we focus on testing the real telemetry behavior. 
I am still not totally happy with the design (e.g. there are some null checks in different places which is a smell for me). I would suggest having two different implementations, e.g. SonarLintProductionTelemetry and another empty one (SonarLintOfflineTelemetry), that would better represent reality. I think we can keep that for later in the sprint, to first make progress on the remaining tickets. WDYT?

damien-urruty-sonarsource: I moved the check for the system property up in the call hierarchy, so that in the tests we focus on testing the real telemetry behavior.
I am still not totally happy with the design (e.g. there are some null checks in different places which is a smell for me). I would suggest having two different implementations, e.g. SonarLintProductionTelemetry and another empty one (SonarLintOfflineTelemetry), that would better represent reality. I think we can keep that for later in the sprint, to first make progress on the remaining tickets. WDYT?

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/SonarLintCorePlugin.java
```diff
@@ -125,9 +126,18 @@ public void startupAsync() {
 
     @Override
     public IStatus run(IProgressMonitor monitor) {
-      telemetry.init();
+      startTelemetry();
       return Status.OK_STATUS;
     }
+
+    private void startTelemetry() {
+      if (SonarLintTelemetry.shouldBeActivated()) {
```
henryju, line 134: I'm not totally sure of the value to move the condition here, compared to keeping everything related to telemetry in the SonarLintTelemetry class.

damien-urruty-sonarsource: It's a first step towards the design I suggested in another comment. My idea is to make explicit that we have 2 different situations, real telemetry in production and disabled telemetry for tests. Ideally I would represent that with 2 different implementations, and the choice between them should be made here IMO. I will try to implement that later this week

PR #235, opened 2020-06-02
Refactor and rename to get rid of 'Server'

henryju:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/wizard/connection/UrlWizardPage.java
```diff
@@ -38,7 +38,7 @@
   private Binding serverUrlTextBinding;
 
   public UrlWizardPage(ServerConnectionModel model) {
-    super("server_url_page", "SonarQube Server URL", model, 2);
+    super("server_url_page", "SonarQube ConnectedEngineFacade URL", model, 2);
```
jblievremont, line 41: Uh, I don't think so :sweat_smile:

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/messages.properties
```diff
@@ -55,7 +55,7 @@ actionUnbind=Unbind
 actionUpdate=Update All Project Bindings
 actionChangeBinding=Change Binding...
 actionEdit=Edit Connection...
-actionNewServer=New Server Connection...
+actionNewServer=New Connection...
 
 # Servers action set
 actionSetNewServer=Create Server
```
jblievremont, line 61: Shouldn't this one be updated too?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintProjectPropertyPage.java
```diff
@@ -135,14 +135,14 @@ private void updateState() {
     Optional<EclipseProjectBinding> projectBinding = getProjectConfig().getProjectBinding();
     if (projectBinding.isPresent()) {
       boundDetails
-        .setText("Bound to the project '" + projectBinding.get().projectKey() + "' on server '" + serverName(projectBinding.get().serverId()) + "'");
+        .setText("Bound to the project '" + projectBinding.get().projectKey() + "' on server '" + serverName(projectBinding.get().connectionId()) + "'");
```
jblievremont, line 138:
```suggestion
        .setText("Bound to the project '" + projectBinding.get().projectKey() + "' on connection '" + serverName(projectBinding.get().connectionId()) + "'");
```

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintProjectPropertyPage.java
```diff
@@ -135,14 +135,14 @@ private void updateState() {
...
       bindLink.setText("<a>Update project binding</a>");
     } else {
       bindLink.setText("<a>Bind this Eclipse project to a SonarQube project</a>");
```
jblievremont, line 141: Found a lonely SonarQube here, might need to be updated to SonarCloud/SonarQube?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintProjectPropertyPage.java
```diff
@@ -135,14 +135,14 @@ private void updateState() {
...
       boundDetails.setText("");
     }
     if (projectBinding.isPresent() && !SonarLintCorePlugin.getServersManager().forProject(getProject()).isPresent()) {
-      addServerLink.setText("<a>Connect to SonarQube server '" + projectBinding.get().serverId() + "'</a>");
+      addServerLink.setText("<a>Connect to SonarQube server '" + projectBinding.get().connectionId() + "'</a>");
```
jblievremont, line 145: Found another lone "SonarQube" here

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/messages.properties
damien-urruty-sonarsource: This `actionSetNewServer` message is used in `NewConnectionWizardAction()` with a `setText()` but the text is immediately replaced with another setText() in `BindingsViewActionProvider`. I think we can safely delete this entry

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AnalyzeConnectedProjectJob.java
```diff
@@ -48,17 +48,17 @@
...
+  public AnalyzeConnectedProjectJob(AnalyzeProjectRequest request, EclipseProjectBinding binding, ConnectedEngineFacade server) {
     super(request);
     this.binding = binding;
     this.server = server;
   }
 
   @Override
   protected ConnectedAnalysisConfiguration prepareAnalysisConfig(Path projectBaseDir, List<ClientInputFile> inputFiles, Map<String, String> mergedExtraProps) {
-    SonarLintLogger.get().debug("Connected mode (using configuration of '" + binding.projectKey() + "' in server '" + binding.serverId() + "')");
+    SonarLintLogger.get().debug("Connected mode (using configuration of '" + binding.projectKey() + "' in server '" + binding.connectionId() + "')");
```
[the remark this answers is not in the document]
damien-urruty-sonarsource: FIxed!

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/wizard/connection/UrlWizardPage.java
damien-urruty-sonarsource: FIxed!

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintProjectPropertyPage.java
damien-urruty-sonarsource: Fixed!

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintProjectPropertyPage.java
damien-urruty-sonarsource: This text is shown when the project is bound but when the binding is 'broken', that is when the connection has been deleted but the project is still bound. If we click on the link, we open the popup to create a new connection, prefilled with the connectionId. 

I changed the text to 'Re-create SonarQube/SonarCloud connection ...'. Because Connect was not the right term here. 

Still I am wondering: this happens probably because the user deleted a connection but didn't unbind the project. When we delete a connection, couldn't we also unbind all the projects bound to this connection ? After informing user of course

org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacadeManagerTest.java
```diff
@@ -31,21 +31,21 @@
...
 /**
  * To have this test pass when launched from the IDE, you have to set -pluginCustomization argument (see pom.xml).
  */
-public class ServersManagerTest {
+public class ConnectedEngineFacadeManagerTest {
 
   private static final ISecurePreferences ROOT_SECURE = SecurePreferencesFactory.getDefault().node(SonarLintCorePlugin.PLUGIN_ID);
   private static final IEclipsePreferences ROOT = InstanceScope.INSTANCE.getNode(SonarLintCorePlugin.PLUGIN_ID);
-  private ServersManager manager;
+  private ConnectedEngineFacadeManager manager;
```
damien-urruty-sonarsource, line 48: What about `ConnectionsManager` ? I usually don't like the term manager on a class because it is vague, in this case it might be ok

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintProjectPropertyPage.java
damien-urruty-sonarsource: This is also the reason why I introduced `ResolvedBinding` in another commit. When a project has a binding, in a lot of places we have to check if the connection is still there, thus the "resolve" operation, which is maybe a poor term.

To summarize, today we have 3 states for a project: unbound, bound to an existing connection, bound to a missing connection. I think we could simplify to keep the 2 first states. If the 3rd happens because of an error, we could remove the binding

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacade.java
```diff
@@ -234,7 +234,8 @@ public String getServerVersion() {
     if (getStorageState() != State.UPDATED) {
       return NEED_UPDATE;
     }
-    return updateStatus.getServerVersion();
+    String version = updateStatus.getServerVersion();
```
henryju, line 237: Here, because of the previous condition, we know that storage state == State.UPDATED, and so getServerVersion() should be non null.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/wizard/connection/ServerConnectionWizard.java
```diff
@@ -298,6 +298,7 @@ public void run(IProgressMonitor monitor) throws InvocationTargetException, Inte
       return false;
     } catch (InterruptedException e) {
       model.setUserOrgs(null);
+      Thread.currentThread().interrupt();
```
henryju, line 301: Eclipse uses InterruptedException to notify of job cancellation. This is not truly a `Thred.interrupt`, so I think it is wrong (and possibly dangerous) to interrupt the current thread.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintProjectPropertyPage.java
henryju: This situation can happen when users share project settings (either in SCM or using workspace mechanics). The project settings contain the binding with a connection id, but the user misses the connection definition. I don't think we should ignore this case.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintProjectPropertyPage.java
```diff
@@ -142,7 +142,7 @@ private void updateState() {
       boundDetails.setText("");
     }
     if (projectBinding.isPresent() && !SonarLintCorePlugin.getServersManager().resolveBinding(getProject()).isPresent()) {
-      addServerLink.setText("<a>Re-bind to SonarQube/SonarCloud connection '" + projectBinding.get().connectionId() + "'</a>");
+      addServerLink.setText("<a>Re-create SonarQube/SonarCloud connection '" + projectBinding.get().connectionId() + "'</a>");
```
henryju, line 145: :+1:

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacade.java
damien-urruty-sonarsource: I made this change because the `GlobalStorageStatus.getServerVersion()` method is annotated with `@CheckForNull` and I got an issue, that's why I added the check. So we can just ignore the issue ?

henryju: Yes, I usually mark it as FP in this case.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/wizard/connection/ServerConnectionWizard.java
damien-urruty-sonarsource: Ok, I was not sure about this one. InterruptedException is normally specific to threads, didn't know Eclipse used them for jobs. So yes I should remove that and just ignore the issue

PR #242, opened 2020-07-02
Rule params ui

kirill-knize-sonarsource:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/util/SonarLintRuleBrowser.java
```diff
@@ -99,6 +99,7 @@ public SonarLintRuleBrowser(Composite parent, boolean useEditorFontSize) {
       JFaceResources.getColorRegistry().addListener(SonarLintRuleBrowser.this);
       this.defaultFont = getDefaultFont();
       JFaceResources.getFontRegistry().addListener(SonarLintRuleBrowser.this);
+
```
jblievremont, line 102: Nitpick: useless diff :wink:

PR #246, opened 2020-07-14
SLE-365 UI bug fixes

kirill-knize-sonarsource:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/RuleParameterPanel.java
```diff
@@ -92,7 +93,7 @@ private void createParamInputs(ScrolledComposite sc) {
     sc.setMinSize(paramInputsContainer.computeSize(SWT.DEFAULT, SWT.DEFAULT));
   }
 
-  private void setDefaultLinkVisibility() {
+  public void setDefaultLinkVisibility() {
```
jblievremont, line 96: Is this change in visibility necessary?

PR #257, opened 2020-09-29
SLE-379 Improve rule description view icon

henryju:


damien-urruty-sonarsource (commented):
The small icon used in the issue context menu also needs to be updated

damien-urruty-sonarsource:
Tested on Fedora, Eclipse 2020-06

damien-urruty-sonarsource (approved):

PR #261, opened 2020-10-08
Make engine facade more thread/null safe

henryju:
There was a possible NPE when an analysis was executed while the engine was restarted


damien-urruty-sonarsource (commented):
Little typo in one of the commit message.

I wonder why it was not working before, because we had the synchronized keyword on the getClient() and stop() methods in the facades, they are the only 2 methods that modify the field, so I don't see how we could have a concurrent access issue.

This withEngine stuff might fix the problem but I think it doesn't help readability.

A question that comes to my mind is: could we avoid having different threads to run analysis ? Couldn't we use a queue instead, maybe per project ?

PR #264, opened 2020-10-16
SLE-391 Fix wizard page title to be dynamic

henryju:


jblievremont (commented):
I could not test it locally (class path issue on ``com.eclipsesource.json.Json``) and there seem to be a failing (flaky?) test, else LGTM.

PR #267, opened 2020-10-27
Try to make concurrent analysis more reliable

henryju:
I'm not totally happy with this change. I did not find a way to run only one analysis when settings are changed.
As a workaround, I'm now ensuring that rules configuration is taken at the last possible moment, and I also tried to limit the workspace locking as much as possible.


jblievremont (approved):
As far as I can tell, it looks like it stabilized the failing test without a visible impact on the perfs.

The one point that worries me a bit is the quick burst of `getJobManager().(begin|end)Rule` for each excluded file, I'm not completely sure what that can mean for overall performance, or how it will play with other plugins.

I think we can already push this to dogfood, and since this is a small self-contained commit, we should be able to revert it quickly if needed.

PR #273, opened 2020-11-12
SLE-396 Fix NPE in Neon

henryju:
I did a test on Neon.3, and we are already using `textEditor.getDocumentProvider().getDocument(textEditor.getEditorInput())` in other places so I'm pretty confident with the fix.


jblievremont (approved):
LGTM, provided the 2020-06 IT axis passes.

Before merging, maybe we could let the community post author try the PR build?

PR #276, opened 2020-12-04
SLE-400 Check if textEditor support annotations

henryju:


jblievremont (approved):
LGTM, weird we did not catch it earlier :thinking:

PR #284, opened 2021-02-18
Open hotspots

henryju:


its/org.sonarlint.eclipse.its/src/org/sonarlint/eclipse/its/utils/CaptureScreenshotAndConsoleOnFailure.java
```diff
@@ -40,16 +37,22 @@ public void evaluate() throws Throwable {
         try {
           base.evaluate();
         } catch (Throwable onHold) {
-          String fileName = constructFilename(description, ".png");
-          new SWTWorkbenchBot().captureScreenshot(fileName);
+          try {
+            String fileName = constructFilename(description, ".png");
+            SWTWorkbenchBot bot = new SWTWorkbenchBot();
+            bot.captureScreenshot(fileName);
```
damien-urruty-sonarsource, line 43: Do we still need this now that we have video recording in the pipeline ?

henryju: Probably not. It may still be useful when you run tests locally, but I don't mind if you prefer removing it (in a later commit).

PR #286, opened 2021-02-18
Create view for security hotspots

henryju:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/hotspots/HotspotsView.java
```diff
@@ -0,0 +1,292 @@
...
+  }
+
+  private Control createVulnerabilityDescriptionControl(Composite parent) {
+    vulnerabilityDescriptionBrowser = new SonarLintWebView(parent, true) {
+
+      @Override
+      protected String body() {
+        ServerHotspot hotspot = (ServerHotspot) hotspotViewer.getStructuredSelection().getFirstElement();
+        if (hotspot == null) {
+          return NO_SECURITY_HOTSPOTS_SELECTED;
```
damien-urruty-sonarsource, line 166: Could this happen ? Couldn't we hide the tabs if no hotspot is selected ?
In IJ, we show the tabs even if the selection is not active, we show them as soon as a hotspot has been opened.

damien-urruty-sonarsource (approved):
:+1:

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/hotspots/HotspotsView.java
henryju: Today it is possible to unselect the hotspot. We could prevent that maybe. We'll see later.

PR #300, opened 2021-03-17
SLE-436 Prevent an exception when getting the file that is being closed

damien-urruty-sonarsource:


henryju (approved):
Maybe we could add an IT that open/close a derived file and check that there is no analysis triggered and no errors.

PR #301, opened 2021-04-13
Introduce GitHub release action

damien-urruty-sonarsource:


.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+        uses: SonarSource/gh-action_LT_release@v2
+
+      - name: Check outputs
+        if: always()
+        run: |
+          echo "${{ steps.sl_release.outputs.releasability }}"
+          echo "${{ steps.sl_release.outputs.release }}"
+      - name: Publish version on P2 update site
+        id: p2_update
+        env:
```
jblievremont, line 48: I wonder if we should add something like this, to only run if the previous steps succeeded:
```suggestion
        if: ${{ success() }}
        env:
```

damien-urruty-sonarsource: If there is no `if` it's already the default value ([source](https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions))

PR #302, opened 2021-04-13
Update the build

henryju:


henryju:
@jblievremont @damien-urruty-sonarsource you might be interested to have a look, FYI

kirill-knize-sonarsource (approved):

PR #306, opened 2021-04-19
SLE-393 Add default position for some views in Java perspective

henryju:


damien-urruty-sonarsource (approved):
:+1: cool improvement! So as you don't target a perspective specifically, I guess the views should be placed the same way in any perspective ? That's even better

kirill-knize-sonarsource (approved):

henryju:
@damien-urruty-sonarsource unfortunately it still targets a specific perspective. I simply grouped the multiple declarations in the same XML element

PR #310, opened 2021-04-23
SLE-438 replace old APIs calls, remove code branches for old versions

kirill-knize-sonarsource:


org.sonarlint.eclipse.m2e/src/org/sonarlint/eclipse/m2e/internal/M2eUtils.java
```diff
@@ -65,9 +65,7 @@ private static IFile toSpecificFile(IFile file) {
     IFile finalFile = file;
     IPath rawLocation = file.getRawLocation();
     if (rawLocation != null) {
-      // TODO use getFileForLocation when we will support only Eclipse 4.6+
-      // IFile moreSpecific = ResourcesPlugin.getWorkspace().getRoot().getFileForLocation(rawLocation);
-      IFile moreSpecific = resourceForLocation(rawLocation);
+      IFile moreSpecific = ResourcesPlugin.getWorkspace().getRoot().getFileForLocation(rawLocation);
```
henryju, line 68: Now you should be able to delete method `resourceForLocation` no?

henryju (changes requested):
I think we should add a lower bound on jface depepdency to prevent installation in unsupported versions

org.sonarlint.eclipse.m2e/src/org/sonarlint/eclipse/m2e/internal/M2eUtils.java
kirill-knize-sonarsource: Yes. And method `resourceFor`. Removed both.

henryju: :+1:

henryju (approved):

PR #313, opened 2021-06-29
Update analyzers for 6.0

damien-urruty-sonarsource:


jblievremont (approved):
Nitpick: it would be great to put the JIRA ticket IDs on each commit, apart from this LGTM 👍🏻

PR #316, opened 2021-06-30
SLE-454 Show workspace name to trusted servers in security hotspot discovery

jblievremont:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/hotspots/SecurityHotspotsHandlerServer.java
```diff
@@ -150,14 +152,19 @@ public void shutdown() {
 
     @Override
     public void handle(ClassicHttpRequest request, ClassicHttpResponse response, HttpContext context) throws HttpException, IOException {
+      boolean trustedServer = Optional.ofNullable(request.getHeader("Origin"))
```
jblievremont, line 155: FYI @damien-urruty-sonarsource I tried to write an IT for this, but `java.net.HttpURLConnection` would not let me change the `Origin` header. We would need to either set a (`sun.*`) system property to allow it, or migrate the tests to e.g Commons HTTP client. Not completely sure it's worth the investment.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/hotspots/SecurityHotspotsHandlerServer.java
```diff
@@ -167,6 +174,10 @@ public void handle(ClassicHttpRequest request, ClassicHttpResponse response, Htt
     }
   }
 
+  private static boolean isTrustedServer(String serverOrigin) {
+    return SonarLintCorePlugin.getServersManager().getServers().stream().anyMatch(s -> s.getHost().contains(serverOrigin));
```
damien-urruty-sonarsource, line 178: I have a doubt about this `contains`. Could we imagine a scenario where an attacker could use an origin that has the registered server URL as a substring ?

damien-urruty-sonarsource (approved):

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/hotspots/SecurityHotspotsHandlerServer.java
jblievremont: Thanks for the comment, I revisited using `startsWith` instead (which more closely matches the actual spec) and commented the code for future reference.

PR #319, opened 2021-07-13
SLE-431 Improve feedback when connection cannot be stored because of secure storage error

damien-urruty-sonarsource:


jblievremont (commented):
Works well on latest Eclipse but I faced issues with Eclipse Luna (oldest supported). Let's discuss.

damien-urruty-sonarsource:
@jblievremont I believe the oldest supported is Neon. I just tested with it and it worked ok

jblievremont (approved):

PR #327, opened 2021-08-30
SLE-463 Improve gitignore checks

damien-urruty-sonarsource:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AbstractAnalyzeProjectJob.java
```diff
@@ -123,7 +123,7 @@ protected IStatus doRun(final IProgressMonitor monitor) {
       FileExclusionsChecker exclusionsChecker = new FileExclusionsChecker(getProject());
       files.forEach(fWithDoc -> {
         ISonarLintFile file = fWithDoc.getFile();
-        if (exclusionsChecker.isExcluded(file, true) || GitUtils.isIgnored(file)) {
+        if (exclusionsChecker.isExcluded(file, true) || file.isScmIgnored()) {
```
henryju, line 126: Would be interesting for investigation to log when the file is excluded due to SCM (see what is done in `exclusionsChecker`)

henryju (approved):
I would add a log then LGTM

kirill-knize-sonarsource (approved):

PR #331, opened 2021-09-15
SLE-470 Display and apply quick fixes

damien-urruty-sonarsource:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/markers/ApplyQuickFixMarkerResolver.java
```diff
@@ -0,0 +1,99 @@
...
+        document.replace(startOffset, endOffset - startOffset, textEdit.getNewText());
+      } catch (Exception e) {
+        SonarLintLogger.get().error("Quick fix location does not exist", e);
+      }
+    }));
+  }
+
+  @Nullable
+  private static ISonarLintFile getFile(IMarker marker) {
+    IFile file = Adapters.adapt(marker.getResource(), IFile.class);
```
henryju, line 77: Why not directly adapting to ISonarLintFile? This would allow to support better the advanced cases when file are not IFiles

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/markers/ApplyQuickFixMarkerResolver.java
```diff
@@ -0,0 +1,99 @@
...
+    return null;
+  }
+
+  private static IDocument getDocument(ISonarLintFile file) {
+    IDocument doc;
+    IEditorPart editorPart = PlatformUtils.findEditor(file);
+    if (editorPart instanceof ITextEditor) {
+      doc = ((ITextEditor) editorPart).getDocumentProvider().getDocument(editorPart.getEditorInput());
+    } else {
+      doc = file.getDocument();
```
henryju, line 90: This part doesn't work. If the file is not open in an editor, this code will return an "in memory" IDocument created from the file content. Changes to this document will have no effect on the file itself.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/markers/ApplyQuickFixMarkerResolver.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+    return doc;
+  }
+
+  @Override
+  public Image getImage() {
+    return SonarLintImages.BALLOON_IMG;
```
henryju, line 97: We need to change the image

henryju (approved):
The main use case works fine, so this PR could probably be merged, and we can add separate tasks to:
- change the icon
- support batch update
- add the quick fix context menu in the SonarLint views
- support applying quick fixes on closed files
- removing the issue when the quick fix is applied (like we do when deactivating a rule)

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/markers/ApplyQuickFixMarkerResolver.java
damien-urruty-sonarsource: I didn't know that I could adapt like this, I think I reused this from somewhere else. Seems to work fine and is simpler so :+1:

damien-urruty-sonarsource: We have the same adaptation code in `DefaultSonarLintProjectAdapter`

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/markers/ApplyQuickFixMarkerResolver.java
damien-urruty-sonarsource: Is it what you refered to in your list as "support applying quick fixes on closed files" ? If so I think I will merge this one and I will add a separate ticket.

henryju: yes

PR #332, opened 2021-09-15
SLE-470 Collect quick fixes usage with telemetry

damien-urruty-sonarsource:


henryju (approved):
LGTM

henryju (changes requested):
We need to update the telemetry sample

PR #333, opened 2021-09-16
Quick fix icon

henryju:


damien-urruty-sonarsource (approved):

org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/jobs/AnalyzeStandaloneProjectJobTest.java
```diff
@@ -200,6 +204,46 @@ public void analyze_report_should_only_triggers_two_events_for_all_marker_operat
...
+      + "\n"
+      + "import java.util.*;\n"
+      + "\n"
+      + "public class FileWithQuickFixes {\n"
+      + "\n"
+      + "  public int foo(boolean a, String foo) {\n"
+      + "    List<String> list = new ArrayList<String>();\n"
+      + "    return 0;\n"
+      + "  }\n"
+      + "}");
```
damien-urruty-sonarsource, line 221: Do we have to repeat the file content here ? Couldn't we load it somehow ?

henryju: I copied blindly what was done in other tests, but in fact it seems we are correctly loading content from disk.

PR #336, opened 2021-09-17
SLE-470 Drop quick fixes that target an old version of the document

damien-urruty-sonarsource:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/EclipseInputFile.java
```diff
@@ -127,4 +130,8 @@ public InputStream inputStream() throws IOException {
     return new ByteArrayInputStream(contents().getBytes(getCharset()));
   }
 
+  public boolean isOlderThan(IDocument document) {
+    return documentModificationStamp < ((IDocumentExtension4) document).getModificationStamp();
```
henryju, line 134: Does it work for closed editors? What is the modificationStamp of a Document created using the `ITextFileBufferManager`?

henryju (approved):
I'm not sure it will work for closed editors, but we can work on it when addressing this topic.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/EclipseInputFile.java
damien-urruty-sonarsource: Nope, that's why a test fails at the moment. Will try to fix that

PR #337, opened 2021-09-21
Quick fixes improvements

henryju:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/markers/ApplyQuickFixMarkerResolver.java
```diff
@@ -99,17 +133,25 @@ private static void applyIn(IDocument document, MarkerQuickFix fix) {
...
+          if (markerPosition.isDeleted()) {
+            // do nothing if position has been deleted
+            return;
+          } else {
+            // use position instead of marker values
+            start = markerPosition.getOffset();
+            end = markerPosition.getOffset() + markerPosition.getLength();
+          }
+        }
+        document.replace(start, end - start, textEdit.getNewText(), document.getModificationStamp() + 1);
```
damien-urruty-sonarsource, line 175: We should maybe have some logs in some cases here. Else it means the application of the fix could have no effect at all (no change in the document and no log)

damien-urruty-sonarsource: Now that I think about it, I think we could probably move/replay some of those checks earlier when the resolvers are contributed. The `IMarkerResolutionGenerator2` is called just when the fixes are about to be displayed, so it could be a good time to check those cases. This would let us not propose a fix instead of proposing it knowing it won't have any effect

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/markers/ApplyQuickFixMarkerResolver.java
```diff
@@ -65,31 +72,58 @@ public void run(IMarker marker) {
...
+    boolean previousValue = projectConfig.isAutoEnabled();
+    projectConfig.setAutoEnabled(false);
+    SonarLintCorePlugin.saveConfig(project, projectConfig);
+    return previousValue;
+  }
+
+  private static void restoreAutoEnabledState(ISonarLintProject project, boolean previousValue) {
+    SonarLintProjectConfiguration projectConfig2 = SonarLintCorePlugin.loadConfig(project);
+    projectConfig2.setAutoEnabled(previousValue);
+    SonarLintCorePlugin.saveConfig(project, projectConfig2);
```
damien-urruty-sonarsource, line 118: I was a bit surprised that we modify real user configuration to achieve this, it seems strange to play with user-facing values to make sure parallel analyses don't happen, I think it's more an internal concern. Why do we need to do this ?

damien-urruty-sonarsource:
Two more comments:

* Should we keep our existing fixes as such ? 
* Should we make sure the quick fixes are displayed in the same order that they were contributed ?

PR #343, opened 2021-09-24
Add ITs for quick fixes

damien-urruty-sonarsource:


its/src/org/sonarlint/eclipse/its/QuickFixesTest.java
```diff
@@ -0,0 +1,186 @@
...
+    ContentAssistant assistant = editor.openQuickFixContentAssistant();
+    assertThat(assistant.getProposals()).element(0).isEqualTo(QUICK_FIX_MESSAGE);
+    assistant.chooseProposal(QUICK_FIX_MESSAGE);
+
+    new WaitUntil(new TextEditorContainsText(editor, EXPECTED_TEXT));
+  }
+
+  @Test
+  public void shouldApplyQuickFixThroughMarkerContextMenu() {
+    // default issue severity is INFO and reddeer supports only ERROR and WARNING
```
henryju, line 79: Can't we use `ProblemType.ALL`?

henryju (approved):

its/src/org/sonarlint/eclipse/its/QuickFixesTest.java
damien-urruty-sonarsource: It wouldn't work for INFO : https://github.com/eclipse/reddeer/blob/master/plugins/org.eclipse.reddeer.eclipse/src/org/eclipse/reddeer/eclipse/ui/views/markers/ProblemsView.java#L46

PR #354, opened 2021-11-03
SLE-479 Fix exception when issue has no 2nd location

jblievremont:


its/src/org/sonarlint/eclipse/its/SecondaryLocationsTest.java
```diff
@@ -46,6 +46,25 @@ public void importProject() {
...
+    onTheFlyView.getItems().get(0).select();
+
+    var flowItems = locationsView.getTree().getItems();
+    assertThat(flowItems).hasSize(1);
+
+    var locationRoot = flowItems.get(0);
+    assertThat(locationRoot.getText()).isEqualTo("No additional locations associated with this issue");
+    locationRoot.select();
+    // SLE-479 No exception should be thrown
+  }
```
damien-urruty-sonarsource, line 66: I prefer to find this kind of info in the git log but I'm fine with it

damien-urruty-sonarsource (approved):
:+1:

PR #357, opened 2021-12-17
Update to SLCORE 8.0

henryju:


henryju:
@kirill-knize-sonarsource this part is only about updating SLCORE, there is no scheduling of the sync every hours. But still I think you can review this part, and I will add the missing job later.

kirill-knize-sonarsource (approved):

PR #366, opened 2022-02-01
SLE-501 Revert to packed bundle and fix URL to Path convertion

henryju:


damien-urruty-sonarsource (approved):
The change looks good but I would add a test specifically around white spaces

PR #369, opened 2022-02-03
SLE-498 Add a bundle to interact with EGit

henryju:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/toolbar/SonarLintToolbarContribution.java
```diff
@@ -81,15 +78,15 @@ public void selectionChanged(IWorkbenchPart part, ISelection selection) {
...
   private void slFileSelected(ISonarLintFile slFile) {
-    updateLabel(VcsService.getServerBranch(slFile.getProject()));
+    updateTooltip(VcsService.getServerBranch(slFile.getProject()));
   }
 
-  private void updateLabel(@Nullable String content) {
-    label.setText(content != null ? content : "");
+  private void updateTooltip(@Nullable String content) {
+    label.setToolTipText(content != null ? content : "");
     label.getParent().getParent().requestLayout();
```
henryju, line 90: @kirill-knize-sonarsource Is it still needed to request a layout update now that we are simply changing the tooltip?

PR #387, opened 2022-03-28
Fix QA

henryju:


damien-urruty-sonarsource (approved):
Very strange bug ! Good catch :+1:

PR #389, opened 2022-04-08
Hardening

henryju:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintGlobalConfiguration.java
```diff
@@ -238,25 +239,37 @@ public static Collection<RuleConfig> readRulesConfig() {
...
+
+    enum Level {
+      @SerializedName("on")
+      ON,
+      @SerializedName("off")
+      OFF
+    }
+
+    Level level = Level.ON;
+    Map<String, String> parameters;
```
kirill-knize-sonarsource, line 252: Could we initialise it with empty map and remove null check below?

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintGlobalConfiguration.java
```diff
@@ -238,25 +239,37 @@ public static Collection<RuleConfig> readRulesConfig() {
...
-      rulesJson.forEach(ruleJson -> {
-        var ruleObject = ruleJson.getValue().asObject();
-        var rule = new RuleConfig(ruleJson.getName(), !RULE_JSON_LEVEL_OFF.equals(ruleObject.getString(RULE_JSON_LEVEL, RULE_JSON_LEVEL_ON)));
-        var ruleParamsValue = ruleObject.get(RULE_JSON_PARAMS);
-        if (ruleParamsValue != null) {
-          var paramsJson = ruleParamsValue.asObject();
-          paramsJson.forEach(p -> rule.getParams().put(p.getName(), p.getValue().asString()));
+      rulesByKey.forEach((key, config) -> {
+        var rule = new RuleConfig(key, RuleConfigGson.Level.OFF != config.level);
+        if (config.parameters != null) {
```
kirill-knize-sonarsource, line 267: This null check could be removed if map was initialised.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/hotspots/SecurityHotspotsHandlerServer.java
```diff
@@ -431,7 +432,9 @@ public ShowHotspotParameters(String projectKey, String hotspotKey, String server
   }
 
   private static class StatusResponse {
+    @SerializedName("ideName")
```
kirill-knize-sonarsource, line 435: This should work without annotation since name in JSON and java field are the same:
https://www.javadoc.io/doc/com.google.code.gson/gson/2.6.2/com/google/gson/annotations/SerializedName.html

kirill-knize-sonarsource (approved):
There are minor remarks but it should work anyway.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/hotspots/SecurityHotspotsHandlerServer.java
```diff
@@ -431,7 +432,9 @@ public ShowHotspotParameters(String projectKey, String hotspotKey, String server
...
     private final String ideName;
+    @SerializedName("description")
```
kirill-knize-sonarsource, line 437: Same here. Should work without annotation.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintGlobalConfiguration.java
henryju: This will produce a different json:
```json
{
  "java:S123": {
    "level": "on",
    "parameters": {
    }
  }
}
```
vs
```json
{
  "java:S123": {
    "level": "on"
  }
}
```

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintGlobalConfiguration.java
henryju: This would produce a different output (see my other comment)

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/hotspots/SecurityHotspotsHandlerServer.java
henryju: Yes, but I like to make things explicit. In case someone refactor the attribute name, it would be a breaking change of the serialized format.

PR #390, opened 2022-04-11
SLE-511 Subscribe for server events

damien-urruty-sonarsource:


its/build.properties
```diff
@@ -1,4 +1,6 @@
 source.. = src/
-output.. = target/classes/
+output.. = target/classes/,\
+           bin/
```
henryju, line 3: Are you sure this change is needed?

its/src/org/sonarlint/eclipse/its/SonarQubeConnectedModeTest.java
```diff
@@ -174,10 +206,55 @@ public void testLocalServerStatusRequest() throws Exception {
...
+    assertThat(defaultEditor.getMarkers())
+      .extracting(Marker::getText, Marker::getLineNumber)
+      .containsOnly(
+        tuple("Replace this use of System.out or System.err by a logger.", 9));
+
+    var qualityProfile = getQualityProfile(JAVA_SIMPLE_PROJECT_KEY, "SonarLint IT Java");
+    deactivateRule(qualityProfile, "S106");
+    Thread.sleep(5000);
+
+    doAndWaitForSonarLintAnalysisJob(() -> {
```
henryju, line 249: Would be great at some point to have analysis automatically triggered when rule is changed.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacade.java
```diff
@@ -364,6 +367,17 @@ public void updateProjectList(IProgressMonitor monitor) {
     });
   }
 
+  @Override
+  public Set<String> getBoundProjectKeys() {
+    return ProjectsProviderUtils.allProjects().stream()
```
henryju, line 372: Maybe you can share part of this code with the following method `getBoundProjects`?

henryju (approved):

its/src/org/sonarlint/eclipse/its/SonarQubeConnectedModeTest.java
damien-urruty-sonarsource: We haven't done it on SLI either, we could consider it later yes

PR #391, opened 2022-04-12
SLE-512 Use JS and HTML analyzers only as an embedded

kirill-knize-sonarsource:


henryju (commented):
Shouldn't we fail if JS or HTML plugins can't be found?

henryju (approved):

PR #395, opened 2022-04-13
Stabilize IT

henryju:


damien-urruty-sonarsource:
Closing as this does not fix ITs

PR #401, opened 2022-04-15
Hardening

henryju:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/ProjectStorageUpdateJob.java
```diff
@@ -40,7 +41,10 @@ public ProjectStorageUpdateJob(String connectionId, String projectKey) {
   protected IStatus run(IProgressMonitor monitor) {
     try {
       var server = SonarLintCorePlugin.getServersManager().findById(connectionId);
-      server.ifPresent(s -> s.updateProjectStorage(projectKey, null, monitor));
+      server.ifPresent(s -> {
+        s.updateProjectStorage(projectKey, null, monitor);
+        s.sync(Set.of(projectKey), monitor);
```
damien-urruty-sonarsource, line 46: Shouldn't we move the `sync` call inside `updateProjectStorage` ? I think the 2 actions should always happen at the same time (the same sequence also happens in `ServerUpdateJob`)

damien-urruty-sonarsource (approved):

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/ProjectStorageUpdateJob.java
henryju: No, because sync should be called on multiple projects keys if possible, not one by one.

PR #403, opened 2022-05-11
BUILD-1578 skip BOM under windows os

julien-carsique-sonarsource:
The BOM is only required under CI and its generation takes ages under Win OS, then let's skip it by default when running under windows family OS.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -256,7 +256,6 @@
           <defaultP2Metadata>false</defaultP2Metadata>
         </configuration>
       </plugin>
-
       <plugin>
         <groupId>org.cyclonedx</groupId>
         <artifactId>cyclonedx-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -302,5 +304,16 @@
...
     </profile>
+    <profile>
+      <id>windows</id>
+      <activation>
+        <os>
+          <family>windows</family>
+        </os>
+      </activation>
+      <properties>
+        <os.windows>true</os.windows>
```
henryju, line 315: Why not simply declaring `<cyclonedx.skip>true</cyclonedx.skip>` here? IMO it makes it easier to understand what this profile is about.

pom.xml
```diff
@@ -302,5 +304,16 @@
         </plugins>
       </build>
     </profile>
+    <profile>
+      <id>windows</id>
```
henryju, line 308:
```suggestion
      <id>skip-cyclonedx-on-windows</id>
```

pom.xml
henryju, line 315:
```suggestion
        <cyclonedx.skip>true</cyclonedx.skip>
```

henryju (commented):
I think we could make the profile more explicit about its use.

henryju (approved):

PR #404, opened 2022-06-06
BUILD-1634 only exec cyclonedx on root project and under CI

julien-carsique-sonarsource:
- BUILD-1634 skip cyclonedx on local build

Execute CycloneDX only when built by CI.
Remove profile skipping CycloneDX under Windows
Workaround conflict between makeAggregateBom + inherited=false and clean triggered by tycho-p2-plugin: exclude bom from clean

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -74,6 +74,7 @@
 
     <gitRepositoryName>sonarlint-eclipse</gitRepositoryName>
     <artifactsToPublish>${project.groupId}:org.sonarlint.eclipse.site:zip,${project.groupId}:${project.artifactId}:json:cyclonedx</artifactsToPublish>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -158,6 +159,25 @@
...
         </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-clean-plugin</artifactId>
+          <version>${version.clean.plugin}</version>
+          <configuration>
+            <excludeDefaultDirectories>true</excludeDefaultDirectories>
+            <filesets>
+              <fileset>
+                <directory>target</directory>
```
henryju, line 170:
```suggestion
                <directory>${project.build.directory}</directory>
```

PR #407, opened 2022-08-02
SLE-545 Update SLCORE for automatic issue sync

henryju:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacade.java
```diff
@@ -620,19 +557,36 @@ public ConnectedEngineFacade setNotificationsDisabled(boolean value) {
 
   public void downloadServerIssues(String projectKey, @Nullable String branchName, IProgressMonitor monitor) {
```
damien-urruty-sonarsource, line 558: Not sure this method is still used (it is already not used on master)

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacade.java
```diff
@@ -214,7 +213,7 @@ private void doWithEngine(Consumer<ConnectedSonarLintEngine> consumer) {
 
   private void reloadProjects(ConnectedSonarLintEngine engine) {
     this.allProjectsByKey.clear();
-    this.allProjectsByKey.putAll(engine.allProjectsByKey());
+    this.allProjectsByKey.putAll(engine.downloadAllProjects(createEndpointParams(), buildClientWithProxyAndCredentials(), null));
```
damien-urruty-sonarsource, line 216: Could this block the UI thread ? Also the `reloadProjects` method is called elsewhere in the class just after downloading all projects, so we could do it twice in a row

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacade.java
```diff
@@ -647,17 +601,37 @@ public Optional<ServerHotspot> getServerHotspot(String hotspotKey, String projec
...
+    for (var projectKey : projectKeys) {
+      var boundProjects = getBoundProjects(projectKey);
+
+      Set<String> activeBranches = new HashSet<>();
+      for (var project : boundProjects) {
+        activeBranches.add(VcsService.getServerBranch(project));
+      }
+
+      activeBranches.forEach(b -> {
+        engine.downloadAllServerIssues(createEndpointParams(), buildClientWithProxyAndCredentials(), projectKey, b,
```
damien-urruty-sonarsource, line 629: Shouldn't we call `syncServerIssues` here instead ? And also `syncServerTaintIssues` ?

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/ProjectStorageUpdateJob.java
```diff
@@ -42,7 +42,7 @@ protected IStatus run(IProgressMonitor monitor) {
     try {
       var server = SonarLintCorePlugin.getServersManager().findById(connectionId);
       server.ifPresent(s -> {
-        s.updateProjectStorage(projectKey, null, monitor);
+        s.updateProjectStorage(projectKey, monitor);
         s.sync(Set.of(projectKey), monitor);
```
damien-urruty-sonarsource, line 46: Shouldn't we also sync issues & taints here ?

damien-urruty-sonarsource (commented):
A few comments/questions. Also:
* there are a few FIXME/TODOs in the code. I suggest we remove them if not important, or fix them or track them in a ticket
* I have not seen sync of issues when changing branch, I could have missed it

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacade.java
henryju: It is used in [AnalyzeConnectedProjectJob](https://github.com/SonarSource/sonarlint-eclipse/blob/3946e656b667db38dfbca4150895dc93b5c9dc6c/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AnalyzeConnectedProjectJob.java#L79)
Seems Eclipse is broken and can not reliably find call sites anymore :(

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/ProjectStorageUpdateJob.java
henryju: The `sync` method is taking care of synchronizing issues as well

damien-urruty-sonarsource (approved):
2 small comments else LGTM

PR #412, opened 2022-09-05
SLE-525 Remove synchronization when using the connected engine

damien-urruty-sonarsource:
Since recent changes in SLCORE, the connected engine can be used by several threads at the same time


henryju (approved):
LGTM!

PR #413, opened 2022-09-06
Extract taint update job

henryju:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/TaintIssuesUpdateJob.java
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  public static void fetchServerTaintIssues(ConnectedEngineFacade engineFacade,
+    ProjectBinding projectBinding,
+    String branchName,
+    ISonarLintFile file, IProgressMonitor monitor) {
+    var filePath = file.getProjectRelativePath();
+
+    try {
+      engineFacade.downloadAllServerTaintIssuesForFile(projectBinding, branchName, filePath, monitor);
```
damien-urruty-sonarsource, line 83: I think for VSCode we call this and also `syncServerTaintIssues` on file open, and maybe we should not

damien-urruty-sonarsource (approved):
LGTM

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/TaintIssuesUpdateJob.java
henryju: I don't think you should. The goal is to stop fetching stuff on file open. Starting from SQ 9.6+, updates should come with SSE. `downloadAllServerTaintIssuesForFile` is kept for older SQ versions (and SC)

PR #414, opened 2022-09-07
SLE-549 Support analysis of JS in YAML files

damien-urruty-sonarsource:
No change in production code as the new YAML language was automatically enabled. We instead list languages to disable.

### DIFF
--- a/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
+++ b/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
@@ -194,6 +194,20 @@ public class StandaloneAnalysisTest extends AbstractSonarLintTest {
         tuple("Refactor the code so this stream pipeline is used.", 13)); // Test that sonar.java.source is set
   }
 
+  @Test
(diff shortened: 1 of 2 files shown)


its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
```diff
@@ -194,6 +194,20 @@ public void shouldAnalyseJava8() {
...
+    new JavaPerspective().open();
+    var rootProject = importExistingProjectIntoWorkspace("js/js-simple", "js-simple");
+
+    openFileAndWaitForAnalysisCompletion(rootProject.getResource("lambda.yaml"));
+
+    var defaultEditor = new DefaultEditor();
+    assertThat(defaultEditor.getMarkers())
+      .extracting(Marker::getText, Marker::getLineNumber)
+      .containsOnly(
+        tuple("Remove the declaration of the unused 'x' variable.", 8)); // Test that sonar.java.source is set
```
henryju, line 208: The comment is a copy/paste that doesn't make sense here.

henryju (approved):
Minor comment then LGTM

PR #422, opened 2022-09-19
SLE-558 Do not assume master is the main branch name

damien-urruty-sonarsource:
The only case where this could happen is if the engine is not started. We throw an error instead


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/engine/connected/ConnectedEngineFacade.java
```diff
@@ -582,8 +582,8 @@ public Optional<ServerHotspot> getServerHotspot(String hotspotKey, String projec
   }
 
   @Override
-  public ProjectBranches getServerBranches(String projectKey) {
-    return withEngine(engine -> engine.getServerBranches(projectKey)).orElse(new ProjectBranches(Set.of("master"), "master"));
+  public Optional<ProjectBranches> getServerBranches(String projectKey) {
+    return withEngine(engine -> engine.getServerBranches(projectKey));
```
henryju, line 586: Why not throwing the exception here?

henryju (approved):

PR #426, opened 2022-12-07
SLE-581 Support Eclipse 2022-12

damien-urruty-sonarsource:


henryju (approved):
Very good refactoring!

PR #431, opened 2023-01-03
SLE-592 Let the core manage the embedded server

damien-urruty-sonarsource:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintProjectConfigurationManager.java
```diff
@@ -49,14 +54,19 @@ public class SonarLintProjectConfigurationManager {
   public static final String P_BINDING_SUGGESTIONS_DISABLED_KEY = "bindingSuggestionsDisabled";
 
   private static final Set<String> BINDING_RELATED_PROPERTIES = Set.of(P_PROJECT_KEY, P_SERVER_ID, P_BINDING_SUGGESTIONS_DISABLED_KEY);
+  private static Map<ISonarLintProject, Collection<IPreferenceChangeListener>> listenersByScopeContext = new HashMap<>();
```
henryju, line 57: I am a bit skeptical about this map. It sounds like a possible memory leak. Entries are never removed (what if for example a project is closed?).

henryju (commented):
Why not keeping the event about the connection been added? The possible problems I see are:
* race conditions (the event is not yet received when the assistCreateConnection response arrive)
* this will trigger auto-binding

The race condition can be mitigated by allowing a short delay on the backend after the assistCreateConnection response.

The assist binding issue can possibly be handled fully on the backend. When an "assist connection creation" is in progress, disable binding suggestions.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintProjectConfigurationManager.java
damien-urruty-sonarsource: Removed, I don't have to unregister/re-register listeners anymore

henryju (approved):

PR #433, opened 2023-01-23
Update parent pom

henryju:


henryju:
@mithil-dobarkar-sonarsource no need to review each file one by one ;) This is more FYI

mithil-dobarkar-sonarsource (approved):

PR #436, opened 2023-02-20
Update QA axes to the new LTS

henryju:


its/pom.xml
```diff
@@ -153,6 +153,7 @@
             <org.eclipse.swt.browser.UseWebKitGTK>true</org.eclipse.swt.browser.UseWebKitGTK>
             <org.eclipse.swt.browser.DefaultType>webkit</org.eclipse.swt.browser.DefaultType>
           </systemProperties>
+          <useJDK>BREE</useJDK>
```
henryju, line 156: This will start ITs with Java 11, because old Eclipse versions (photon) doesn't work with Java 17. And in general I think it is better to continue testing with Java 11.

damien-urruty-sonarsource (approved):

PR #439, opened 2023-02-22
SLE-603 Update okhttp and use it as an OSGi bundle

henryju:


target-platforms/commons.target
```diff
@@ -1,42 +1,70 @@
...
+				  <type>jar</type>
+			  </dependency>
+		  </dependencies>
+		  <instructions><![CDATA[
+Bundle-Name:           OkHttp
+version:               ${version_cleanup;${mvnVersion}}
+Bundle-SymbolicName:   com.squareup.okhttp3
+Bundle-Version:        ${version}
+Export-Package:        okhttp3;version="${version}";-noimport:=true
+Import-Package:        okio;version="3.0.0",kotlin.*;version="1.6.20"
```
damien-urruty-sonarsource, line 56: Why is this needed ?

damien-urruty-sonarsource (approved):
One question, else LGTM

target-platforms/commons.target
henryju: Because okhttp is not an OSGi library (while okio is...):
image

henryju: Full OSGi support is coming in version 5: https://square.github.io/okhttp/changelogs/changelog/#version-500-alpha1

PR #451, opened 2023-04-26
SLE-614: Sonar property based on project settings

thahnen:
## Summary

The new property "sonar.java.enablePreview" will be set automatically based on the specific project settings. If the project setting is unavailable or disabled (default or on purpose by user), the property value will be set to "False", otherwise to "True".

### DIFF
--- a/org.sonarlint.eclipse.jdt/src/org/sonarlint/eclipse/jdt/internal/JdtUtils.java
+++ b/org.sonarlint.eclipse.jdt/src/org/sonarlint/eclipse/jdt/internal/JdtUtils.java
@@ -99,8 +99,8 @@ public class JdtUtils {
     context.setAnalysisProperty("sonar.java.source", javaSource);
     context.setAnalysisProperty("sonar.java.target", javaTarget);
     
-    var javaPreview = Optional.ofNullable(javaProject.getOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, true)).orElse("disabled");
-    context.setAnalysisProperty("sonar.java.enablePreview", javaPreview.equalsIgnoreCase("enabled") ? "true" : "false");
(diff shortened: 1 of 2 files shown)


org.sonarlint.eclipse.jdt/src/org/sonarlint/eclipse/jdt/internal/JdtUtils.java
```diff
@@ -97,6 +97,10 @@ public void configureJavaProject(IJavaProject javaProject, IPreAnalysisContext c
 
     context.setAnalysisProperty("sonar.java.source", javaSource);
     context.setAnalysisProperty("sonar.java.target", javaTarget);
+    
+    var javaPreview = javaProject.getOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, true);
+    context.setAnalysisProperty("sonar.java.enablePreview",
+      javaPreview != null && javaPreview.equalsIgnoreCase("enabled") ? "True" : "False");
```
henryju, line 103: Boolean analyzer properties should be lowercase. Also, I think this way of writing code is safer:
```suggestion
    var javaPreview = Optional.ofNullable(javaProject.getOption(JavaCore.COMPILER_PB_ENABLE_PREVIEW_FEATURES, true)).orElse(false);
   context.setAnalysisProperty("sonar.java.enablePreview", javaPreview.toString());
```

henryju (changes requested):

org.sonarlint.eclipse.jdt/src/org/sonarlint/eclipse/jdt/internal/JdtUtils.java
thahnen: I altered your suggestion a bit to work (the `getOption` returns a String), but other than that everything as suggested.

henryju (approved):

PR #453, opened 2023-04-28
SLE-609: Smart notifactions via sonarlint-core

thahnen:
## Summary

The smart notifications feature is enabled to be handled by sonarlint-core. Remove dead / unused code on the Eclipse side.

## Test plan

Test done manually and everything passed as excepted, see PDF as attachment: [SLE-609 - Manual Test Plan.pdf](https://github.com/SonarSource/sonarlint-eclipse/files/11351833/SLE-609.-.Manual.Test.Plan.pdf)
The only "issue" found can be seen in the PDF on Screenshot 2: When removing the last project from the Eclipse workspace, the "SonarLint Bindings" view does not update and still shows the connection and the project removed before. But I think this is not related to the current ticket.


henryju:
> When removing the last project from the Eclipse workspace, the "SonarLint Bindings" view does not update and still shows the connection and the project removed before.

@thahnen  Can you please create a separate ticket for this bug please?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintUiPlugin.java
```diff
@@ -80,11 +76,9 @@ public class SonarLintUiPlugin extends AbstractUIPlugin {
   @Nullable
   private SonarLintConsole console;
 
-  private ListenerFactory listenerFactory;
-
   private static final WindowOpenCloseListener WINDOW_OPEN_CLOSE_LISTENER = new WindowOpenCloseListener();
   private static final SonarLintPostBuildListener SONARLINT_POST_BUILD_LISTENER = new SonarLintPostBuildListener();
-  private static final SonarLintProjectEventListener SONARLINT_PROJECT_EVENT_LISTENER = new SonarLintProjectEventListener();
+  private static final SonarLintVcsCleanupListener SONARLINT_PROJECT_EVENT_LISTENER = new SonarLintVcsCleanupListener();
```
henryju, line 81: The static variable should be renamed too. And after more thinking, a better name is maybe:
```suggestion
  private static final SonarLintVcsCacheCleaner SONARLINT_VCS_CACHE_CLEANER = new SonarLintVcsCacheCleaner();
```

henryju (approved):

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintUiPlugin.java
thahnen: Done

PR #454, opened 2023-04-28
SLE-615: Whitelist enabled languages

thahnen:
## Summary

Languages are added to `sonarlint-core`, so every new language must be disabled. This change simpliefies the process and disables the the addition of an unsupported language by accedentially updating `sonarlint-core` without checking for new languages.


henryju (approved):
LGTM

PR #455, opened 2023-05-04
SLE-597: WIP

thahnen:
WIP


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/RulesConfigurationPart.java
```diff
@@ -208,19 +211,32 @@ public int compare(Viewer viewer, Object e1, Object e2) {
...
-        paramPanel = new RuleParameterPanel(paramPanelParent, SWT.NONE, wrapper.ruleDetails, wrapper.ruleConfig);
-        paramPanel.setLayoutData(new GridData(SWT.FILL, SWT.FILL, true, true));
+      var ruleKey = wrapper.ruleDetails.getKey();
+
+      try {
+        ruleBrowser.updateRule(SonarLintBackendService.get()
+          .getBackend()
+          .getRulesService()
+          .getStandaloneRuleDetails(new GetStandaloneRuleDescriptionParams(ruleKey))
+          .get());
```
henryju, line 226: This is probably interesting to do this in a background job, in order to not freeze the UI

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/util/SonarLintRuleBrowser.java
```diff
@@ -35,51 +36,59 @@
...
 import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneRuleDetails;
 import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneRuleParam;
+import org.sonarsource.sonarlint.core.clientapi.backend.rules.GetEffectiveRuleDetailsResponse;
+import org.sonarsource.sonarlint.core.clientapi.backend.rules.GetStandaloneRuleDescriptionResponse;
 
+/** "SonarLint Rule Description" view content (monolithic / educational rule description) */
 public class SonarLintRuleBrowser extends SonarLintWebView {
-
-  private RuleDetails ruleDetails;
+  private Either<GetEffectiveRuleDetailsResponse, GetStandaloneRuleDescriptionResponse> ruleDetailsResponse;
```
henryju, line 46: I would try to avoid leaking the "Response" objects everywhere. This might be done later in other subtasks, but I think this `SonarLintRuleBrowser` component should only deal with `Either<RuleMonolithicDescriptionDto, RuleSplitDescriptionDto> description`

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/views/RuleDescriptionWebView.java
```diff
@@ -148,17 +150,26 @@ private void showRuleDescription(IMarker element) {
...
-    if (resolvedBindingOpt.isPresent()) {
-      new AsyncDisplayRuleDescriptionJob(project, resolvedBindingOpt.get(), ruleKey, browser).schedule();
-    } else {
-      var ruleDetails = SonarLintCorePlugin.getInstance().getDefaultSonarLintClientFacade().getRuleDescription(ruleKey);
-      browser.updateRule(ruleDetails);
+    try {
+      var project = Adapters.adapt(element.getResource(), ISonarLintIssuable.class).getProject();
+      var resolvedBindingOpt = SonarLintCorePlugin.getServersManager().resolveBinding(project);
+      if (resolvedBindingOpt.isPresent()) {
+        var configurationScopeId = resolvedBindingOpt.get().getProjectBinding().projectKey();
```
henryju, line 157: Configuration scope ID is not the projectKey, but the "identifier" of the Eclipse project. The backend is already aware of the binding, so it will deal with that on its own.
I suggest to create a method in the `SonarLintBackendService` to encapsulate the conversion ISonarLintProject to configScopeId.

PR #457, opened 2023-05-04
SLE-619: Switch to new RulesService

thahnen:
## Summary

Switched to the new RulesService, therefore this PR is more or less a WIP. Updated the background jobs for loading the SonarLintRuleBrowser content to differ based on project / global configuration level.

## Issues

The only issue I'm currently facing is the following: Having a project opened which is bound to my local SQ instance, when analyzing a file of that project and the local SQ instance is down, the `SonarLintBackendService.getEffectiveRuleDetails` will hang in the background job as it awaits its own timeout. It doesn't know for now, that the connection might not be available at all.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/SonarLintBackendService.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/SonarLintBackendService.java
@@ -38,6 +38,7 @@ import org.sonarlint.eclipse.core.internal.engine.connected.IConnectedEngineFaca
 import org.sonarlint.eclipse.core.internal.jobs.GlobalLogOutput;
 import org.sonarlint.eclipse.core.internal.preferences.SonarLintGlobalConfiguration;
 import org.sonarlint.eclipse.core.internal.utils.SonarLintUtils;
+import org.sonarlint.eclipse.core.resource.ISonarLintProject;
 import org.sonarsource.sonarlint.core.SonarLintBackendImpl;
 import org.sonarsource.sonarlint.core.clientapi.SonarLintBackend;
 import org.sonarsource.sonarlint.core.clientapi.SonarLintClient;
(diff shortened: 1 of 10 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/SonarLintBackendService.java
```diff
@@ -158,6 +163,41 @@ private static String getIdeName() {
     return ideName;
   }
 
+  /**
```
henryju, line 166: I am not convinced of the value of the javadoc here. This is not really an API, and most infos can be guessed from method signature

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/job/DisplayProjectRuleDescriptionJob.java
```diff
@@ -19,42 +19,43 @@
...
-  private final ResolvedBinding binding;
+/** Update "web browser" view for the project rule description (maybe context based on connection) in a separate thread */
+public class DisplayProjectRuleDescriptionJob extends AbstractSonarProjectJob {
+  private final ISonarLintProject project;
   private final String ruleKey;
+  private final String contextKey;
   private final SonarLintRuleBrowser browser;
 
-  public AsyncDisplayRuleDescriptionJob(ISonarLintProject project, ResolvedBinding binding, String ruleKey, SonarLintRuleBrowser browser) {
+  public DisplayProjectRuleDescriptionJob(ISonarLintProject project, String ruleKey, String contextKey, SonarLintRuleBrowser browser) {
```
henryju, line 40: You can mark `contextKey` as Nullable

henryju (approved):
Minor comments but overall LGTM. See my previous draft PR to fix the IT.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/SonarLintBackendService.java
thahnen: Yeah I agree, shrunk the comment to only contain a short summary. For me a short summary is necessary, as the method names differ so much, I would play cluedo the next time I encounter that after a long period trying to figure out why the first one contains `Description` and the other one `Details` 👍

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/job/DisplayProjectRuleDescriptionJob.java
thahnen: Done, also in `SonarLintBackendService.getEffectiveRuleDetails`.

thahnen:
Some ITS are currently failing as I removed the `SonarLintRuleBrowser` HTML content for now, they shall work again when we fix the UI!

PR #458, opened 2023-05-04
 SLE-623 Store rule description context on markers 

henryju:


thahnen (approved):
Looks good, in the `RuleDescriptionWebView` it can be directly added on top of my changes without changing anything, neat!

PR #461, opened 2023-05-09
SLE-620 Display rule header with native components

henryju:


thahnen (changes requested):
The only differnce between the old and the new PR: The `SonarLintRuleBrowser` should have been moved from `internal/util` to `internal/rule`. In this branch the class is available two times and shown as new class added, not moved via Git. Ccan you please remove the old class at `interal/util/SonarLintRuleBrowser.java`?

Otherwise everything else is LGTM!

thahnen (approved):
LGTM

PR #462, opened 2023-05-09
SLE-621 Create tabs to display rules with sections

henryju:


henryju:
@nquinquenel FYI ;)

nquinquenel (approved):
Splendid!

PR #467, opened 2023-06-05
SLE-641: Code snippets in SourceViewer

thahnen:
## Summary

Basic syntax highlighting based on the Eclipse plugins providing language support. Currently only implemented for CDT / JDT. Support for other plugins like [PDT](https://sonarsource.atlassian.net/browse/SLE-635) or [PyDev](https://sonarsource.atlassian.net/browse/SLE-636) etc. might come after that, starting with the other official languages and then some other ones developed externally but open source.

## Issues

Reacting to theme changes automatically to reload the `SourceViewer` source code snippets does not work yet. Also changes to the `META-INF/MANIFEST.MF` files of the plugins regarding imported / required packages has to be checked.

## Testing

Currently unit tests and integration tests are still failing as the current testing was only done manually from _Eclipse RDP_ itself and with the packaged plugin inside _Eclipse CPP_.
Also testing was only done on macOS so far!

### DIFF
--- a/org.sonarlint.eclipse.jdt/src/org/sonarlint/eclipse/jdt/internal/JavaProjectConfiguratorExtension.java
+++ b/org.sonarlint.eclipse.jdt/src/org/sonarlint/eclipse/jdt/internal/JavaProjectConfiguratorExtension.java
@@ -20,6 +20,7 @@
 package org.sonarlint.eclipse.jdt.internal;
 
 import java.util.Collections;
+import java.util.EnumSet;
 import java.util.Optional;
 import java.util.Set;
 import org.eclipse.core.resources.IFile;
@@ -27,6 +28,7 @@ import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.resources.IProject;
 import org.eclipse.core.runtime.IProgressMonitor;
 import org.eclipse.jdt.annotation.Nullable;
+import org.eclipse.jface.text.IDocumentPartitioner;
 import org.eclipse.jface.text.source.SourceViewerConfiguration;
 import org.sonarlint.eclipse.core.analysis.IAnalysisConfigurator;
 import org.sonarlint.eclipse.core.analysis.IFileTypeProvider;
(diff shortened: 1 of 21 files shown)


org.sonarlint.eclipse.jdt/src/org/sonarlint/eclipse/jdt/internal/JavaProjectConfiguratorExtension.java
```diff
@@ -73,7 +75,7 @@ private static boolean isClassPresentAtRuntime(String className) {
   @Override
   public Set<Language> whitelistedLanguages() {
     if (isJdtPresent()) {
-      return Collections.singleton(Language.JAVA);
+      return EnumSet.of(Language.JAVA, Language.JSP);
```
henryju, line 78: Maybe this change would have been better in a separate commit

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDescriptionPanel.java
```diff
@@ -49,76 +54,151 @@
...
+    IDocumentPartitioner documentPartitionerNullable = null;
+    for (var configurationProvider : configurationProviders) {
+      var documentPartitionerOptional = configurationProvider.documentPartitioner(languageKey);
+      if (documentPartitionerOptional.isPresent()) {
+        documentPartitionerNullable = documentPartitionerOptional.get();
+        break;
+      }
+    }
+
+    var snippetElement = new SourceViewer(parent, null, SWT.BORDER);
```
henryju, line 182: I think we should remove the border at some point :)

henryju (approved):
Minor feedback, I will merge in the common branch so that we can move on

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDescriptionPanel.java
thahnen: Yeah, agree. Was only done for seeing the `SourceViewer` boundaries locally and when they change.

henryju: In fact, I realized there is a border in the Eclipse formater settings, so in the end maybe we could keep it.

PR #469, opened 2023-06-09
Syntax highlighting

henryju:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDescriptionPanel.java
```diff
@@ -179,13 +179,14 @@ private static void createSourceViewer(String html, Composite parent, String lan
       }
     }
 
-    var snippetElement = new SourceViewer(parent, null, SWT.BORDER);
+    var snippetElement = new SourceViewer(parent, null, SWT.BORDER | SWT.H_SCROLL);
     var gridData = new GridData();
     gridData.horizontalAlignment = SWT.FILL;
     gridData.grabExcessHorizontalSpace = true;
+    gridData.horizontalIndent = 10;
```
thahnen, line 186: Is the indentation intended? Or was it added just for testing?

thahnen (approved):
LGTM

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDescriptionPanel.java
henryju: This is intended, I find it looks better like this. WDYT?

PR #470, opened 2023-06-12
BUILD-2809 try gh-action_release 5.0.8

julien-carsique-sonarsource:
Validate future 5.0.8 release of gh-action_release that includes CloudFront invalidation of the S3 cache.


henryju (approved):
Let's see during the next release

PR #477, opened 2023-06-19
SLE-658: sonarlint-core 8.19 update

thahnen:
.


henryju (changes requested):
Build is failing, I think you missed some of the changes

henryju (approved):
LGTM, maybe cleanup the commits: keep everything related to SLCORE update in one commit, and another commit for the fix of ITs that are unrelated?

PR #484, opened 2023-07-04
SLE-665: Issue marker heritage

thahnen:
Issues now show their heritage, meaning that in connected mode the issues present on the server can be safely distinguished from the ones only available locally.


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintImages.java
```diff
@@ -88,22 +88,31 @@ private SonarLintImages() {
...
+   *  
+   *  @param marker used for finding the project connection mode
+   *  @param severity issue severity
+   *  @param type issue type
+   *  @return composite image if found, null otherwise
    */
   @Nullable
-  public static Image getIssueImage(String severity, @Nullable String type) {
-    var key = severity + "/" + type;
+  public static Image getIssueImage(@Nullable IMarker marker, @Nullable String markerServerKey, String severity, @Nullable String type) {
```
henryju, line 99: Passing the marker here is maybe creating a bit too much coupling. This class is about providing images, and in this case you simply need the "mode"

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/markers/MarkerUtils.java
```diff
@@ -53,6 +55,10 @@ public final class MarkerUtils {
 
   public static final Set<String> SONARLINT_PRIMARY_MARKER_IDS = Set.of(
     SonarLintCorePlugin.MARKER_ON_THE_FLY_ID, SonarLintCorePlugin.MARKER_REPORT_ID, SonarLintCorePlugin.MARKER_TAINT_ID);
+  
+  public static final String SONAR_MARKER_MODE_STANDALONE = "standalone";
```
henryju, line 59: I wonder if an enum would not be more appropriate here

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/markers/MarkerUtils.java
```diff
@@ -66,6 +72,25 @@ public static void updateAllSonarMarkerSeverity() throws CoreException {
...
+   *  
+   *  @param markerId for the marker <-> project connection
+   *  @param markerServerKey marker information from the connection, null if not on server
+   *  @return markers' corresponding project connection mode
+   */
+  public static String getProjectConnectionMode(IMarker marker, @Nullable String markerServerKey) {
+    var bindingOptional = SonarLintCorePlugin.getServersManager()
+      .resolveBinding(Adapters.adapt(marker.getResource().getProject(), ISonarLintProject.class));
+    if (bindingOptional.isEmpty() || markerServerKey == null) {
+      return SONAR_MARKER_MODE_STANDALONE;
```
henryju, line 87: This is a bit confusing to have "standalone" when the project is actually bound. Maybe the names are not correct. We want to know if the issue is matched with a server issue.
So maybe have an enum with values:
```
enum FindingMatchingStatus {
  NOT_MATCHED,
  MATCHED_WITH_SQ,
  MATCHED_WITH_SC
}
```

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintImages.java
henryju: After my previous comment, I think we could rename "mode" to "matchingStatus".

henryju (changes requested):
The code looks good overall, I am just challenging the naming and suggest to avoid "leaking" marker in the SonarLintImages utility class.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/markers/MarkerUtils.java
thahnen: Done

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/markers/MarkerUtils.java
thahnen: Done

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintImages.java
```diff
@@ -66,10 +66,10 @@ public final class SonarLintImages {
   public static final Image SONARQUBE_SERVER_ICON_IMG = createImage("logo/sonarqube-16px.png"); //$NON-NLS-1$
   public static final Image SONARQUBE_PROJECT_ICON_IMG = createImage("project-16x16.png"); //$NON-NLS-1$
   public static final Image SONARCLOUD_SERVER_ICON_IMG = createImage("logo/sonarcloud-16px.png"); //$NON-NLS-1$
-  public static final Image SONARLINT_ICON_IMG = createImage("logo/sonarlint_logo.png"); //$NON-NLS-1$
   public static final Image IMG_SONARQUBE_LOGO = createImage("logo/sonarqube-black-256px.png"); //$NON-NLS-1$
   public static final Image IMG_SONARCLOUD_LOGO = createImage("logo/sonarcloud-black-256px.png"); //$NON-NLS-1$
 
+  public static final ImageDescriptor STANDALONE_16 = createImageDescriptor("standalone-16.png"); //$NON-NLS-1$
```
henryju, line 72:
```suggestion
  public static final ImageDescriptor NOT_MATCHED_16 = createImageDescriptor("standalone-16.png"); //$NON-NLS-1$
```

henryju (approved):
Minor comment else LGTM!

PR #485, opened 2023-07-04
SLE-675 Check permission to administer issues

henryju:


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/command/MarkAsResolvedCommand.java
```diff
@@ -55,10 +64,51 @@ private static Optional<ResolvedBinding> getBinding(IMarker marker) {
...
+    var checkJob = new Job("Check user permissions for administering issues") {
+
+      private CheckStatusChangePermittedResponse result;
+      private ResolvedBinding resolvedBinding;
+
+      @Override
+      protected IStatus run(IProgressMonitor monitor) {
+        var issueKey = selectedMarker.getAttribute(MarkerUtils.SONAR_MARKER_SERVER_ISSUE_KEY_ATTR, null);
+        if (issueKey == null) {
+          return Status.error("No issue key for this marker");
```
thahnen, line 77: I think this error message might be misleading. I would clarify and say that for this specific marker, no corresponding issue on the server could be found.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/command/MarkAsResolvedCommand.java
```diff
@@ -55,10 +64,51 @@ private static Optional<ResolvedBinding> getBinding(IMarker marker) {
   }
 
   @Override
-  protected void execute(IMarker selectedMarker) {
-    var msg = new MessageBox(Display.getDefault().getActiveShell());
-    msg.setText("TODO");
-    msg.open();
+  protected void execute(IMarker selectedMarker, IWorkbenchWindow window) {
+    var checkJob = new Job("Check user permissions for administering issues") {
```
thahnen, line 68: Maybe `Check user permissions for setting the issue resolution` or something like this?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/command/MarkAsResolvedCommand.java
```diff
@@ -55,10 +64,51 @@ private static Optional<ResolvedBinding> getBinding(IMarker marker) {
...
+
+      @Override
+      protected IStatus run(IProgressMonitor monitor) {
+        var issueKey = selectedMarker.getAttribute(MarkerUtils.SONAR_MARKER_SERVER_ISSUE_KEY_ATTR, null);
+        if (issueKey == null) {
+          return Status.error("No issue key for this marker");
+        }
+        var binding = getBinding(selectedMarker);
+        if (binding.isPresent()) {
+          resolvedBinding = binding.get();
```
thahnen, line 81: This class variable can be removed entirely, or am I mistaken?

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/JobUtils.java
```diff
@@ -48,6 +51,61 @@ public static <T> T waitForFuture(IProgressMonitor monitor, CompletableFuture<T>
     }
   }
 
+  /**
+   * Run something after the job is done, regardless of result.
+   * Important: call job.schedule() after calling this method, NOT before.
+   */
+  public static void scheduleAfter(Job job, Runnable runnable) {
```
thahnen, line 58: Neat this refactoring!

thahnen (approved):
LGTM, only some comments ^^

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/command/MarkAsResolvedCommand.java
henryju: I don't think we need to be too specific here, as this error should "never" occurs, as the action is normally only triggered for markers having an issue key. This is just in case of race condition (the issue key would be removed between the moment the user clicked on the context menu and now).

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/command/MarkAsResolvedCommand.java
henryju: It is passed to the downstream method after the job has completed

PR #488, opened 2023-07-07
SLE-667: Dialog for muting issues

thahnen:
## Summary
This adds a dialog for the muting issues action based on the IntelliJ implementation. Allowed issue transitions are added dynamically based on the permissions check prior.

Also the `CompletableFuture` instances from the `IssueService` method invocations are currently not functional as I did not want to change anything by mistake, I'm mostly testing with SLLS.

## Testing
Not yet tested but in the end when everything is ready we should add integration tests both for muting an issue and a taint vulnerability!


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/command/MarkAsResolvedCommand.java
```diff
@@ -58,24 +67,37 @@ public void updateElement(UIElement element, Map parameters) {
...
+      return;
+    }
+    
+    String markerType;
+    try {
+      markerType = selectedMarker.getType();
+    } catch (CoreException err) {
+      window.getShell().getDisplay()
+      .asyncExec(() -> MessageDialog.openError(window.getShell(), "Mark Issue as Resolved, marker type not available",
+        err.getMessage()));
```
henryju, line 91: Stacktrace is lost, maybe you can log it?

henryju (approved):
Minor feedback then LGTM

PR #495, opened 2023-07-12
SLE-626: Node.js on macOS

thahnen:
## Summary
Automatic detection of Node.js on macOS does not work when added to $PATH from the user shell configuration.

## Testing
Additional changes to ITs finally make the tests run on macOS. This issue of file associations was fixed with this commit.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/NodeJsManager.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/NodeJsManager.java
@@ -19,9 +19,11 @@
  */
 package org.sonarlint.eclipse.core.internal.utils;
 
+import java.nio.file.InvalidPathException;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.Objects;
+import java.util.Optional;
 import org.eclipse.jdt.annotation.Nullable;
 import org.sonarlint.eclipse.core.internal.SonarLintCorePlugin;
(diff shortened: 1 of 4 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/NodeJsManager.java
```diff
@@ -65,6 +78,11 @@ private synchronized void initNodeIfNeeded() {
       this.nodeInit = true;
       this.nodeJsPath = helper.getNodeJsPath();
       this.nodeJsVersion = helper.getNodeJsVersion();
+      
+      if (Util.isMac()) {
+        SonarLintLogger.get().warn(
+          "Automatically detecting Node.js on macOS does not work when added to $PATH from user shell configuration!");
```
henryju, line 84: I am not convinced this log will be really useful. Most users won't look at the console anyway. Shouldn't we log only when `nodeJsPath` is null?

henryju (approved):

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/NodeJsManager.java
thahnen: I've added a warning when it is null but kept the macOS warning. Some users might find it useful when debugging themselves. Otherwise, it will be a hint for us if we have to provide support regarding Node.js at some point 😄

PR #496, opened 2023-07-12
SLE-678: await when accessing markers

thahnen:
Using `await().unitlAsserted(...)` for tests checking for issue markers that are dependent on an analysis.


thahnen:
@henryju I don't get why the ITs are now failing. The usage of `await().untilAsserted(...)` should have prevented that, as the failure comes from the UI command after that check.

thahnen:
This PR is too old, I will reopen the topic sometime later in the future

PR #503, opened 2023-07-18
SLE-616: Update binding view based on project

thahnen:
When a project is closed (or reopened) or deleted, the SonarLint Bindings view should be updated to not include irrelevant projects which might lead to errors when trying to change the binding of it. This does not include importing a project.

### DIFF
--- a/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
+++ b/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
@@ -271,7 +271,7 @@ public class BindingsView extends CommonNavigator {
       } else if (event.getDelta() != null) {
         try {
           event.getDelta().accept((IResourceDelta delta) -> {
-            if (delta.getKind() == IResourceDelta.CHANGED) {
+            if (delta.getKind() == IResourceDelta.CHANGED && (delta.getResource().getType() & IResource.ROOT) != 0) {
               // INFO: With adding "IResourceDelta.ADDED" to the mask we can also update the binding view when
               //       importing a project. Therefore the user can see if a newly imported project is already
               //       correctly binded.
--- a/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
+++ b/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
@@ -74,7 +74,9 @@ public class BindingsView extends CommonNavigator {
 
   private IConnectedEngineFacadeLifecycleListener serverResourceListener;
   private IConnectedEngineFacadeListener serverListener;
+  
   private IResourceChangeListener projectListener;
+  private volatile boolean shouldRefresh = false;
 
   public BindingsView() {
     super();
@@ -263,29 +265,38 @@ public class BindingsView extends CommonNavigator {
     
     // add listener for when project is opened / closed / deleted
     projectListener = event -> {
-      if (event.getBuildKind() == IResourceChangeEvent.PRE_CLOSE
-        || event.getBuildKind() == IResourceChangeEvent.PRE_DELETE) {
+      if (event.getType() == IResourceChangeEvent.PRE_CLOSE
+        || event.getType() == IResourceChangeEvent.PRE_DELETE) {
+        shouldRefresh = true;
+      } else if (event.getDelta() != null) {
+        try {
+          event.getDelta().accept((IResourceDelta delta) -> {
+            if (delta.getKind() == IResourceDelta.CHANGED) {
+              // INFO: With adding "IResourceDelta.ADDED" to the mask we can also update the binding view when
+              //       importing a project. Therefore the user can see if a newly imported project is already
+              //       correctly binded.
+              for (var child : delta.getAffectedChildren(IResourceDelta.CHANGED)) {
+                var resource = child.getResource();
+                
+                if (((resource.getType() & IResource.PROJECT) != 0) 
+                  && resource.getProject().isOpen()
+                  && ((child.getFlags() & IResourceDelta.OPEN) != 0)) {
+                  shouldRefresh = true;
+                  return true;
+                }
+              }
+            }
+            
+            return false;
+          });
+        } catch (CoreException err) {
+          SonarLintLogger.get().error("Error while updating the SonarLint Bindings after opening a project", err);
+        }
+      }
+      
+      if (shouldRefresh) {
         refreshView();
-        return;
-      }
-      
-      if (event.getDelta() == null) {
-        return;
-      }
-      
-      try {
-        event.getDelta().accept((IResourceDelta delta) -> {
-          var rsrc = delta.getResource();
-          if (((rsrc.getType() & IResource.PROJECT) != 0) 
-            && rsrc.getProject().isOpen()
-            && delta.getKind() == IResourceDelta.CHANGED
-            && ((delta.getFlags() & IResourceDelta.OPEN) != 0)) {
-            refreshView();
-          }
-          return true;
-        });
-      } catch (CoreException err) {
-        SonarLintLogger.get().error("Error while updating the SonarLint Bindings after opening a project", err);
+        shouldRefresh = false;
       }
     };
     ResourcesPlugin.getWorkspace().addResourceChangeListener(projectListener);
--- a/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
+++ b/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
@@ -19,6 +19,7 @@
  */
 package org.sonarlint.eclipse.ui.internal.binding;
 
+import java.util.concurrent.atomic.AtomicBoolean;
 import org.eclipse.core.resources.IResource;
 import org.eclipse.core.resources.IResourceChangeEvent;
 import org.eclipse.core.resources.IResourceChangeListener;
@@ -74,9 +75,7 @@ public class BindingsView extends CommonNavigator {
 
   private IConnectedEngineFacadeLifecycleListener serverResourceListener;
   private IConnectedEngineFacadeListener serverListener;
-  
   private IResourceChangeListener projectListener;
-  private volatile boolean shouldRefresh = false;
 
   public BindingsView() {
     super();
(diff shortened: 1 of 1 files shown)


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/AbstractProjectChangeListener.java
```diff
@@ -0,0 +1,57 @@
...
+
+import org.eclipse.core.resources.IProject;
+import org.eclipse.core.resources.IResourceChangeEvent;
+import org.eclipse.core.resources.IResourceChangeListener;
+import org.eclipse.core.resources.IWorkspace;
+import org.eclipse.core.resources.ResourcesPlugin;
+import org.eclipse.jdt.annotation.Nullable;
+
+public abstract class AbstractProjectChangeListener implements IResourceChangeListener {
+  protected IWorkspace workspace;
```
henryju, line 30: Not sure storing a reference on the Workspace is a good idea

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/AbstractProjectChangeListener.java
```diff
@@ -0,0 +1,57 @@
...
+  
+  protected AbstractProjectChangeListener(IProjectChangeHandler handler) {
+    this.workspace = ResourcesPlugin.getWorkspace();
+    this.handler = handler;
+  }
+  
+  @Override
+  public void resourceChanged(IResourceChangeEvent event) {
+    var rsrc = event.getResource();
+    if (rsrc instanceof IProject) {
```
henryju, line 42: We should avoid relying on IProject (think about our Cobol friends). Instead, you can try to adapt the resource to an ISonarLint project. If it adapts, then work with this ISonarLintProject later on.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
```diff
@@ -128,7 +135,7 @@ public void widgetSelected(SelectionEvent e) {
           commonViewer, commonViewer.getNavigatorContentService());
 
         var menuManager = new MenuManager("#PopupMenu");
-        menuManager.addMenuListener(mgr -> {
+        menuManager.addMenuListener((IMenuManager mgr) -> {
```
henryju, line 138: Is this change required?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
```diff
@@ -243,7 +251,7 @@ public void connectionRemoved(IConnectedEngineFacade server) {
     };
     SonarLintCorePlugin.getServersManager().addServerLifecycleListener(serverResourceListener);
 
-    serverListener = facade -> {
+    serverListener = (IConnectedEngineFacade facade) -> {
```
henryju, line 254: Is this change required?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
```diff
@@ -252,7 +260,17 @@ public void connectionRemoved(IConnectedEngineFacade server) {
...
+    
+    // add listener for when project is closed / deleted
+    AbstractProjectChangeListener.IProjectChangeHandler handler = (IProject project) -> {
+      refreshConnectionState();
+      for (var server : SonarLintCorePlugin.getServersManager().getServers()) {
+        refreshConnectionContent(server);
+      }
+    };
+    
+    projectCloseDeleteListener = new ProjectCloseDeleteListener(handler);
```
henryju, line 272: I am not totally convinced by the need to have 2 different listeners, while in practice you have the same handler in both cases.
Also this AbstractProjectChangeListener seems a bit over-engineered to me. Why not simply having `BindingViews`  implementing IResourceChangeListener? I think this si a common practice in Eclipse codebase.

henryju (changes requested):

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
thahnen: No, done

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/AbstractProjectChangeListener.java
thahnen: I removed the project entirely as it does not work with `ISonarLintProject`

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/AbstractProjectChangeListener.java
thahnen: Gone, done

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
```diff
@@ -252,7 +260,42 @@ public void connectionRemoved(IConnectedEngineFacade server) {
     for (var server : SonarLintCorePlugin.getServersManager().getServers()) {
       server.addConnectedEngineListener(serverListener);
     }
-
+    
+    // add listener for when project is opened / closed / deleted
+    projectListener = event -> {
+      if (event.getBuildKind() == IResourceChangeEvent.PRE_CLOSE
+        || event.getBuildKind() == IResourceChangeEvent.PRE_DELETE) {
+        refreshView();
```
henryju, line 268: Won't it fire every time a resource is deleted, like a file or a folder?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
```diff
@@ -252,7 +260,42 @@ public void connectionRemoved(IConnectedEngineFacade server) {
...
+      }
+      
+      try {
+        event.getDelta().accept((IResourceDelta delta) -> {
+          var rsrc = delta.getResource();
+          if (((rsrc.getType() & IResource.PROJECT) != 0) 
+            && rsrc.getProject().isOpen()
+            && delta.getKind() == IResourceDelta.CHANGED
+            && ((delta.getFlags() & IResourceDelta.OPEN) != 0)) {
+            refreshView();
```
henryju, line 283: Shouldn't you return "false" after that, to prevent multiple calls to "refresh". In fact, it might be more robust to collect a boolean "shouldRefresh", and only call "refreshView" outside the delta visitor.

henryju (changes requested):

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
thahnen: No the event is only for projects

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
thahnen: Done

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
```diff
@@ -252,7 +262,51 @@ public void connectionRemoved(IConnectedEngineFacade server) {
...
+        || event.getType() == IResourceChangeEvent.PRE_DELETE) {
+        shouldRefresh = true;
+      } else if (event.getDelta() != null) {
+        try {
+          event.getDelta().accept((IResourceDelta delta) -> {
+            if (delta.getKind() == IResourceDelta.CHANGED) {
+              // INFO: With adding "IResourceDelta.ADDED" to the mask we can also update the binding view when
+              //       importing a project. Therefore the user can see if a newly imported project is already
+              //       correctly binded.
+              for (var child : delta.getAffectedChildren(IResourceDelta.CHANGED)) {
```
thahnen, line 278: FYI: The delta resource is the Workspace and therefore we have to check if the direct children have changed. We specifically only look for changed projects (excluding hidden ones like "External Plug-in Libraries").

thahnen:
Adapting to ISonarLintProject did not work at all, tested that multiple times on multiple levels. If this very feature does not work on Cobol editors, I have to say it is a risk I'm willing to take. This way it can be a QOL improvement at least for the other developers.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
```diff
@@ -252,7 +262,51 @@ public void connectionRemoved(IConnectedEngineFacade server) {
     for (var server : SonarLintCorePlugin.getServersManager().getServers()) {
       server.addConnectedEngineListener(serverListener);
     }
-
+    
+    // add listener for when project is opened / closed / deleted
+    projectListener = event -> {
+      if (event.getType() == IResourceChangeEvent.PRE_CLOSE
```
henryju, line 268: Why not declaring the `shouldRefresh` variable here, instead of risking a race condition?

henryju (approved):
I would move the `shouldRefresh` field to a local variable, then LGTM

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/BindingsView.java
thahnen: Cannot be done 😞 When used as well in `IResourceDelta.accept(...)` in line 273+ it cannot be compiled with the issue: "Variable shouldRefresh used in lambda expression should be final or effectively final".
The ideal resolution would be to use a class variable and in case of the listener being delegated by the workspace to different threads (which I cannot foresee) to use `volatile`.

@henryju regarding a race condition: From manual testing, it seems that the events are processed via a queue but I'm not sure if it is really the case. I can move it to an `AtomicBoolean` maybe?

henryju: Yeah, I think a local AtomicBoolean is better in my opinion

thahnen: Done

PR #506, opened 2023-07-20
SLE-659 SLE-633 Fix branch and taint synchronization

henryju:
As discussed, I am not super happy about all those changes. I would have preferred to move most of the logic to the backend, but it will take longer than expected. So this is a kind of workaround in the meantime.


thahnen (approved):
LGTM!

I see your pain-points and I also like to see it at one point in time to be not part of SLE anymore. Do we have a ticket for making the final changes to SLCORE? If not, please create one 😃

henryju:
The ticket is here: https://sonarsource.atlassian.net/browse/SLCORE-496

PR #507, opened 2023-07-21
SLE-674 Don't notify because of disabled language

henryju:


thahnen (approved):
LGTM and thanks for adding a unit test!

PR #510, opened 2023-08-08
SLE-694: Remove deprecated methods

thahnen:
*SonarLint for Eclipse* does not rely on the deprecated methods from `StandaloneSonarLintEngine` from *SonarLint Core* anymore.

### DIFF
--- a/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/RuleParameterPanel.java
+++ b/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/RuleParameterPanel.java
@@ -35,18 +35,17 @@ import org.eclipse.swt.widgets.Text;
 import org.sonarlint.eclipse.core.SonarLintLogger;
 import org.sonarlint.eclipse.core.internal.preferences.RuleConfig;
 import org.sonarlint.eclipse.core.internal.utils.StringUtils;
-import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneRuleDetails;
-import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneRuleParam;
-import org.sonarsource.sonarlint.core.client.api.standalone.StandaloneRuleParamType;
+import org.sonarsource.sonarlint.core.clientapi.backend.rules.RuleDefinitionDto;
+import org.sonarsource.sonarlint.core.clientapi.backend.rules.RuleParamDefinitionDto;
+import org.sonarsource.sonarlint.core.clientapi.backend.rules.RuleParamType;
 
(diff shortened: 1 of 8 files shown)


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDetailsPanel.java
```diff
@@ -110,31 +111,38 @@ private void updateScrollCompositeMinSize() {
...
-    updateHtmlDescription(getStandaloneRuleDescriptionResponse.getDescription(), ruleDefinition.getLanguage().getLanguageKey());
+      updateHtmlDescription(getStandaloneRuleDescriptionResponse.getDescription(), ruleDefinition.getLanguage().getLanguageKey());
 
-    requestLayout();
-    updateScrollCompositeMinSize();
+      requestLayout();
+      updateScrollCompositeMinSize();
+    } catch (SWTException ignored) {
+      // There might be a race condition between the background job running late and the view already being closed
+    }
```
damien-urruty-sonarsource, line 127: I'm a bit puzzled by this, it looks strange to me that we have to catch such an exception. Isn't there anything we could do to prevent it from happening?
Same remark for the method below

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/RuleParameterPanel.java
```diff
@@ -87,26 +86,26 @@ private void createParamInputs(ScrolledComposite sc) {
     var layout = new GridLayout();
     layout.numColumns = 2;
     paramInputsContainer.setLayout(layout);
-
-    selectedRuleMetadata.paramDetails().forEach(it -> addParamInput(paramInputsContainer, it));
+
```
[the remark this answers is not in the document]
thahnen: I dunno I've checked again and there is actually no formatting issue when opening the file locally 😮

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDetailsPanel.java
thahnen: We populate the rule content in a background job triggered e.g. when switching the rule outside of the UI thread. When it takes some time (e.g. on the slow Cirrus CI build and I got it on my private machine) and in the meantime, the view / Settings panel is closed, then the exception is thrown regarding the parent composite already being disposed of.
It is not caught in the background job in its try-catch block as we run the methods via `Display.getDefault().syncExec(() -> RuleDetailsPanel.updateRule(...))`.

PR #512, opened 2023-08-10
SLE-689 / SLE-691: Core changes for rule descriptions

thahnen:
## Summary

Changes to the rule description inside the specific view and the settings. The fake data is here to stay while testing and no official analyzer is available for dogfooding!


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleHeaderPanel.java
```diff
@@ -28,49 +29,101 @@
...
-  private final Label ruleTypeLabel;
-  private final Label ruleSeverityIcon;
-  private final Label ruleSeverityLabel;
+  private final Label label1;
+  private final Label label2;
+  private final Label label3;
+  private final Label label4;
+  private final Label label5;
+  private final Label label6;
+  private final Label label7;
```
damien-urruty-sonarsource, line 57: I think fields named like this are super hard to follow, and it's very easy to mix them up. If the panels have different structures, why not introduce 2 different classes? One `LegacyRuleHeaderPanel` (that will be deleted at some point) and this class that would be used for the new CCT?

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintImages.java
```diff
@@ -133,6 +130,11 @@ public static Image getSeverityImage(IssueSeverity severity) {
   public static Image getTypeImage(RuleType type) {
     return createImage("type/" + type.name().toLowerCase(Locale.ENGLISH) + ".png");
   }
+  
+  @Nullable
+  public static Image getImpactImage(ImpactSeverity impact) {
+    return createImage("impact/" + impact.name() + ".png");
+  }
```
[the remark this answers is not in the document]
thahnen: No, we have it cached in the `ImageRegistry`: The first time we call `getImpactImage` it is created, and the other times only retrieved.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDetailsPanel.java
```diff
@@ -111,32 +117,32 @@ private void updateScrollCompositeMinSize() {
...
+      
+      ruleHeaderPanel.updateRule(ruleInformation);
 
-      updateHtmlDescription(details.getDescription(), details.getLanguage().getLanguageKey());
-      updateParameters(details);
+      updateHtmlDescription(description, ruleInformation.getLanguage().getLanguageKey());
+      
+      if (ruleInformation instanceof EffectiveRuleDetailsDto) {
+        updateParameters((EffectiveRuleDetailsDto) ruleInformation);
+      }
```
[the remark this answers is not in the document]
thahnen: Sadly no because we have to run `requestLayout()` and `updateScrollCompositeMinSize()` afterward, therefore moving it to the other method would move this out of the shared method as well. Additionally, the try-catch block also applies to this `updateParameters(...)` call and so it would have to be moved to the separate methods as well.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleHeaderPanel.java
thahnen: I see the point and I have no idea how to name them better. The issue with having two classes is that we would have to dispose of and create a new one from scratch every time we call `RuleDetailsPanel#updateRule(...)` based on the new CCT or the old CCT.
The **Rule Description view** is sadly already quite slow and flickery due to the disposing of the `RuleDescriptionPanel` every time we update it and I favor stable usability over maintainability in this case. Yes such a naming is hard to follow but with the description and it being a minor and simple class I think it is manageable.
WDYT?

damien-urruty-sonarsource: Another idea instead of adding/removing the view each time would be to just hide the one not in use: you could create the 2 classes, always add them to the tree and `setVisible` only one of them based on the CCT to use. 
The other argument is that the day we remove the legacy CCT, and I know it's not for tomorrow, it will be simpler to remove the legacy class.
I agree it's a tradeoff, I would still prefer maintainability and readability in this case over efficiency

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDetailsPanel.java
```diff
@@ -109,40 +99,52 @@ private void updateScrollCompositeMinSize() {
...
+      ruleHeaderPanel.dispose();
+    }
+    
+    var attributeOptional = ruleInformation.getCleanCodeAttribute();
+    var impacts = ruleInformation.getDefaultImpacts();
+    
+    if (attributeOptional.isPresent() && !impacts.isEmpty()) {
+      ruleHeaderPanel = new RuleHeaderPanel(scrolledContent);
+    } else {
+      ruleHeaderPanel = new LegacyRuleHeaderPanel(scrolledContent);
```
nquinquenel, line 143: I'm not familiar with Eclipse code but if this method is purely for the rule panel (not the issue panel), then we only want to show the new CCT. In any case, we will always have the new analyzer in the upcoming release, so we should expect only new CCT for rules.

The only case with old CCT is with SQ <10.2, which means it's only for issues in connected mode

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleHeaderPanel.java
thahnen: Resolved but was only possible by disposing of the `AbstractRuleHeaderPanel`. But this did not cause the big deal I was anticipating so no further decrease of performance on loading a rule description.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleDetailsPanel.java
thahnen: No, everything in Eclipse is shared 😄 This is reused in the **Rule Description view** (for the issues) and the **Rules Configuration settings panel** (for the rules).

thahnen:
@damien-urruty-sonarsource the two code smells will be fixed on the parent branch with SLE-695

its/pom.xml
```diff
@@ -159,6 +159,7 @@
           <forkedProcessTimeoutInSeconds>1800</forkedProcessTimeoutInSeconds>
           <redirectTestOutputToFile>true</redirectTestOutputToFile>
           <systemProperties>
+			<target.platform>${target.platform}</target.platform>
```
damien-urruty-sonarsource, line 162: The indentation looks strange here

its/src/org/sonarlint/eclipse/its/reddeer/views/RuleDescriptionView.java
```diff
@@ -50,7 +50,7 @@ public Label getRuleSeverity() {
   }
 
   public Label getRuleKey() {
-    return new DefaultLabel(getCTabItem(), 5);
+    return new DefaultLabel(getCTabItem(), 8);
```
damien-urruty-sonarsource, line 53: What does this 8 represent? Maybe you could extract it as a constant and give it a meaningful name?

damien-urruty-sonarsource (approved):
LGTM

its/src/org/sonarlint/eclipse/its/reddeer/views/RuleDescriptionView.java
thahnen: Yeah, this is on my agenda to clarify that magic number when I'll be implementing all the ITs / UTs. It is currently switched to 8 because we have 8 labels for the Rule Description view header (where clean code attribute / impacts / rule key live). 5 will be for the IT with the connection to an old SQ with the legacy header shown.

PR #519, opened 2023-08-21
SLE-692: Education for the new Clean Code Taxonomy

thahnen:
## Summary

Skeletons for the upcoming changes regarding the documentation. Most of the skeletons must be only filled with content or extended.


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleHeaderPanel.java
```diff
@@ -59,23 +60,45 @@ public void updateRule(AbstractRuleDto ruleInformation) {
...
+      secondImpactSeverityIcon.setImage(SonarLintImages.getImpactImage(secondImpactSeverity));
+      secondImpactSeverityIcon.setToolTipText(createToolTip(secondImpactSeverity, secondSoftwareQuality));
+      secondSoftwareQualityLabel.setText(secondSoftwareQuality);
+      // TODO: Change to SLCORE provided tooltip!;
+      //firstSoftwareQualityLabel.setToolTipText(keys.get(1).getToolTip());
       
       if (keys.size() > 2) {
-        thirdImpactSeverityIcon.setImage(SonarLintImages.getImpactImage(impacts.get(keys.get(2))));
-        thirdSoftwareQualityLabel.setText(clean(impacts.get(keys.get(2)).getDisplayLabel()));
+        var thirdImpactSeverity = impacts.get(keys.get(2));
```
damien-urruty-sonarsource, line 87: The three blocks here have a lot in common, do you think you could factorize them somehow? Maybe have a separate `ImpactPanel` class?

damien-urruty-sonarsource (commented):
I am a little dubious about this change, for 3 reasons:
* I think now that we have a beautiful site it should be the go-to place for all documentation and FAQ. The doc would be easier to update/fix than having it embedded in the product, and accessible by everyone without installing the product
* The links we had before (hotspots, taints) were pointing to exactly the right information, now they open a dialog page where the information is mixed with other things not relevant to the user journey
* As it is now, the descriptions of the FAQ entries are rather short but if they become longer or if we keep adding questions, I fear it will become quickly hard to read. And Eclipse is sometimes limited with UI widgets we can use
* and a last one as a bonus, it is not a FAQ as the last 2 items are not questions :)

I know you probably discussed this with the right people, just wanted to add my perspective

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/rule/RuleHeaderPanel.java
thahnen: Done

thahnen:
@damien-urruty-sonarsource after further discussion we concluded to not yet include the documentation information in the settings.

its/src/org/sonarlint/eclipse/its/reddeer/views/RuleDescriptionView.java
```diff
@@ -50,7 +50,7 @@ public Label getRuleSeverity() {
   }
 
   public Label getRuleKey() {
-    return new DefaultLabel(getCTabItem(), 8);
+    return new DefaultLabel(getCTabItem(), 5);
```
thahnen, line 53: As we now have the same number of direct child elements in the `RuleDescriptionView` this change done earlier is reverted

org.sonarlint.eclipse.ui/icons/full/marker_resolution16/info.png
thahnen, line 1: This icon is the same present in SLI / SLVSCODE

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/command/HelpOnIconsCommand.java
```diff
@@ -0,0 +1,35 @@
...
+
+import org.eclipse.core.resources.IMarker;
+import org.eclipse.ui.IWorkbenchWindow;
+import org.sonarlint.eclipse.ui.internal.util.BrowserUtils;
+
+/** When the user wants to know what the issue / rule icons mean (and why they differ between old / new CCT) */
+public class HelpOnIconsCommand extends AbstractIssueCommand {
+  @Override
+  protected void execute(IMarker selectedMarker, IWorkbenchWindow window) {
+    // TODO: Change when documentation is ready
```
thahnen, line 30: Kevin is aware and I'll await the final documentation containing information on the icons

org.sonarlint.eclipse.ui/plugin.xml
```diff
@@ -1099,6 +1120,10 @@
           class="org.sonarlint.eclipse.ui.internal.command.ShowRuleDescriptionCommand"
           commandId="org.sonarlint.eclipse.ui.internal.command.ShowRuleDescriptionCommand">
     </handler>
+    <handler
+          class="org.sonarlint.eclipse.ui.internal.command.HelpOnIconsCommand"
+          commandId="org.sonarlint.eclipse.ui.internal.command.HelpOnIconsCommand">
+    </handler>
```
damien-urruty-sonarsource, line 1126: Should we make them visible only when necessary, e.g. when clicking on a marker? What I did on the [other PR](https://github.com/SonarSource/sonarlint-eclipse/pull/520) might help

damien-urruty-sonarsource (approved):
LGTM

PR #520, opened 2023-08-23
SLE-697 Show commands only when relevant

damien-urruty-sonarsource:


org.sonarlint.eclipse.ui/plugin.xml
```diff
@@ -579,15 +579,6 @@
                  icon="icons/full/marker_resolution16/disablerule.png"
                  style="push"
                  mnemonic="u">
-          <visibleWhen>
```
thahnen, line 582: Why was it removed here? I'm currently not able to see the reason behind that change.

thahnen (approved):
LGTM!

org.sonarlint.eclipse.ui/plugin.xml
damien-urruty-sonarsource: Because I moved it under the handler declaration (2nd addition in the PR)

PR #521, opened 2023-08-23
SLE-709 Fix blank contexts in rule description

damien-urruty-sonarsource:


thahnen (approved):
LGTM! Even though not perfect (as we discussed earlier) this is also working way better on macOS ^^

PR #525, opened 2023-08-31
SLE-717 Clarify the format of the test file pattern property

henryju:


thahnen (approved):
LGTM! But can you please add a link to the **SonarLint -> File Exclusions** preference page like "_Learn more_" linking to the docs page [HERE](https://docs.sonarsource.com/sonarlint/eclipse/using-sonarlint/file-exclusions/)?

henryju:
I am not sure File Exclusions are using the same format, unfortunately. I asked Kevin to create a new dedicated page in the doc. We could add a link to it when it is ready.

thahnen:
Ah, I was assuming we are using the same logic for patterns on both preference pages.

PR #527, opened 2023-09-12
SLE-722 Add a SonarLint menu in the status bar

henryju:


damien-urruty-sonarsource (approved):
LGTM, maybe at some point you could introduce some UI tests to verify the widget

PR #528, opened 2023-09-14
SLE-724 Display the matched branch in the binding view

henryju:


damien-urruty-sonarsource (approved):
LGTM, I'm surprised there was no test to update, might be good to add an IT at some point

PR #534, opened 2023-09-26
SLE-735 Fix menu visibility on Eclipse 2023-09+

henryju:


thahnen (approved):
LGTM! Good catch, Damien and I changed the visibility stuff on the handlers before the last release just because we watched the context menu entries throwing exceptions (but just tiny ones) when available in the wrong situation.

PR #535, opened 2023-09-28
SLE-725 Display the new code period in the binding view

henryju:


thahnen (approved):
LGTM, this is nice!

We might have to think about when certain situations occur (like the branch being changed, there is a New Code Definition server-sent event in the future, the server directory changing), that we might want to have a listener in the future where we can easily let the `SonarLint Bindings view` reload its content?

PR #539, opened 2023-10-04
Avoid resourceChanged computation for IncrementalProjectBuilder.CLEAN_BUILD

RoiSoleil:
As mentionned in the IncrementalProjectBuilder.CLEAN_BUILD documentation 'Resource deltas are not applicable for this kind of build.'

### DIFF
--- a/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintPostBuildListener.java
+++ b/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintPostBuildListener.java
@@ -25,6 +25,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.stream.Collectors;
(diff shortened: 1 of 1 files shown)


henryju:
Hi @RoiSoleil 

Can you describe the issue you are fixing with this patch please? Is this a performance optimization, or was it creating a bug?

RoiSoleil:
Sorry, yes this is a performance optimisation. I did noticed that when i was doing a "Maven update" is was way longer when SonarLint is enabled. It was because the SonarLintPostBuildListener#resourceChanged was doing a lot of computation.

thahnen:
Hi @RoiSoleil,

I'm in order for us to let the build system test the changes and do some quick investigations if it might have other impacts, I'd like you to change the target branch to `external/RoiSoleil` and then I'll create a new PR to master. None of the information will be lost and you will still be the one doing the commit but this way we can integrate it in our build system.

Best,
Tobias

RoiSoleil:
It's done ;)

PR #546, opened 2023-10-11
MMF-3246: Anticipated issue transitions

thahnen:
## Summary

Connected to SonarQube 10.2+ there is now the possibility to mute local issues before they appear in the next analysis. This way the user does not have to await said analysis if it is clear that the issue will be marked as resolved / won't fix / false positive / ... later on anyway.

For already resolved issues the user will have the opportunity to re-open them from the IDE, therefore resolved issues should be visible in the IDE if the user wants to.

## Testing

Due to the integration tests always failing on Cirrus CI, this new feature was tested manually.


org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/jobs/AnalyzeStandaloneProjectJobTest.java
```diff
@@ -108,7 +111,7 @@ public void analyzeWithRuleParameters() throws Exception {
 
     var underTest = new AnalyzeStandaloneProjectJob(new AnalyzeProjectRequest(slProject, List.of(fileToAnalyze), TriggerType.EDITOR_CHANGE));
     underTest.schedule();
-    assertThat(underTest.join(10_000, new NullProgressMonitor())).isTrue();
+    assertThat(underTest.join(100_000, new NullProgressMonitor())).isTrue();
```
damien-urruty-sonarsource, line 114: Do you really need to wait 100s?

damien-urruty-sonarsource (approved):
I don't see anything wrong on the implementation side. But I'm concerned about the tests marked as ignored (I counted 7). I would suggest to investigate why they are failing rather sooner, cause we could have a risk to introduce regressions

org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/jobs/AnalyzeStandaloneProjectJobTest.java
thahnen: It will wait, at most, 100 seconds. The reason for that change is that running the unit test locally most often takes longer than 10 seconds, sometimes only 11 seconds, and sometimes more than a minute because the test workspace needs to synchronize before starting tests.

thahnen:
Some tests being ignored on the CI is a bummer, I agree but there is just no capacity currently to investigate it. The tests run flawlessly locally but due to the CI being more or less a black box and we only run them locally on the latest Eclipse version, the investigation is very painful.
I currently un-ignore the tests locally when testing new features, I did the same for this MMF ("manual testing").

PR #547, opened 2023-10-13
BUILD-3911: Upload JUnit Plug-in Test logs

thahnen:
When running the "unit" tests we face the issue that on failures prior to running the tests (setting up environment) or after (tearing down environment) the log is not printed to the console but to the specific workspace file.

@matemoln while just forwarding the file content to the console was possible, it was very messy as the log can be quite huge.


thahnen:
@matemoln just ignore the failing QA for now, I'm fixing it on a different branch as it is a flaky IT

matemoln (approved):

PR #554, opened 2023-10-31
SLE-747: IDE to server path matching failure

thahnen:
Handle the corner case in which a file with issues cannot be matched to server issues because the file itself does not comply with the server path.

### DIFF
--- a/org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/tracking/ProjectIssueTrackerTest.java
+++ b/org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/tracking/ProjectIssueTrackerTest.java
@@ -20,14 +20,22 @@
 package org.sonarlint.eclipse.core.internal.tracking;
 
 import java.io.IOException;
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
 import java.util.List;
+import org.eclipse.core.internal.resources.File;
+import org.eclipse.core.runtime.NullProgressMonitor;
 import org.junit.Before;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
+import org.sonarlint.eclipse.core.SonarLintLogger;
+import org.sonarlint.eclipse.core.internal.LogListener;
 import org.sonarlint.eclipse.core.resource.ISonarLintFile;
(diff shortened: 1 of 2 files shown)


org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/core/internal/tracking/ProjectIssueTrackerTest.java
```diff
@@ -207,4 +230,17 @@ public void should_match_local_issues_by_text_range_hash() {
...
 
+  /** Corner case in which the IDE path cannot be converted to server path */
+  @Test
+  public void test_trackWithServerIssues_path_null() {
+    try {
+      underTest.trackWithServerIssues(new ProjectBinding("testProject", "/", "/"), List.of(file1), false, null);
+    } catch(NullPointerException ignored) {
+      // INFO: Ignore the failure on the job AFTERWARDS, this test only checks for the corner case!
+      // INFO: The NPE is coming from 'JobUtils.waitForFuture' because 'monitor' is null, not from the SLCORE method!
+    }
```
damien-urruty-sonarsource, line 241: Couldn't you create a dummy monitor instead of catching the NPE?

thahnen: I could, but then it fails with an NPE when running the Job at the point it is trying to get the `configScopeId` because I don't mock the resource and project of the `ISonarLintFile` in the test class.
"Failing fast", in this case, is the laziest but best option to test the corner case, in my opinion.

damien-urruty-sonarsource: IMHO it's never a good option to catch NPE, there is even a Sonar [rule](https://rules.sonarsource.com/java/RSPEC-1696/) about that, because it can potentially hide other problems. What if one day a regression is introduced, a "real" NPE is thrown elsewhere but silently caught here?
The usage of mocks is another smell to me, but that's a different debate, and it would imply rewriting the tests.
So I would suggest to find another solution, maybe relying on mocks

thahnen: The test is fixed using a mock

PR #555, opened 2023-10-31
SLE-745: Strip extra analyzer properties

thahnen:
We don't want to have extra analyzer properties containing some form of white spaces (which can cause issues), therefore we strip everything when saving them.


damien-urruty-sonarsource (approved):
Maybe a test somewhere would be useful?

PR #557, opened 2023-11-02
SLE-636: Python Syntax Highlighting

thahnen:
## Summary

In the Rule Description view/settings page we offer syntax highlighting only when the PyDev plug-in is available.

## Testing

Testing it via an integration test forced me to increase the timeout time for integration tests as we more often hit the upper boundary of 30 minutes due to Cirrus CI flaky performance.


its/pom.xml
```diff
@@ -149,14 +149,15 @@
...
-          <forkedProcessTimeoutInSeconds>600</forkedProcessTimeoutInSeconds>
           <trimStackTrace>false</trimStackTrace>
           <useUIHarness>true</useUIHarness>
           <useUIThread>false</useUIThread>
           <appArgLine>-eclipse.keyring target/keyring -eclipse.password secure-storage-password.txt -pluginCustomization
             ${project.basedir}/plugin_customization.ini</appArgLine>
-          <forkedProcessTimeoutInSeconds>1800</forkedProcessTimeoutInSeconds>
+          
+          <!-- Kill test JVM if tests take more than 45 minutes (2700 seconds) to finish -->
+          <forkedProcessTimeoutInSeconds>2700</forkedProcessTimeoutInSeconds>
```
nquinquenel, line 159: What's the reason for going from 10 minutes timeout to 45 minutes?

its/target-platforms/ibuilds.target
```diff
@@ -16,6 +16,10 @@
       <repository location="https://download.eclipse.org/reddeer/releases/latest" />
       <unit id="org.eclipse.reddeer.eclipse.feature.feature.group" version="0.0.0" />
     </location>
+    <location includeAllPlatforms="false" includeConfigurePhase="false" includeMode="planner" includeSource="true" type="InstallableUnit">
+      <unit id="org.python.pydev.feature.feature.group" version="0.0.0" />
+      <repository location="http://pydev.sourceforge.net/pydev_update_site/11.0.3" />
```
nquinquenel, line 21: Why is this `pydev_update_site/11.0.3` here and in other places it's `pydev_update_site/10.2.1`?

nquinquenel (approved):
Not an expert on Eclipse but it looks similar to what was done for Java 👍🏻

its/pom.xml
thahnen: The diff contained two `forkedProcessTimeoutInSeconds` statements 😄 The 10min one was an old copy paste one I noticed, the second statement (you can also see in this preview) is the old 30min one.

its/target-platforms/ibuilds.target
thahnen: We have 3 target platforms for ITs: **Oldest**, **Latest**, **iBuilds**.

We use **Latest** also in Eclipse for development, **iBuilds** is what is the latest on Eclipse IDE and the plug-in side as we normally don't build with what is either just released or still in development. **Oldest** is of course the oldest stuff we support (or try to support).

nquinquenel: Make sense, thanks 👍🏻

PR #560, opened 2023-11-06
NO-JIRA: Remove Eclipse user survey

thahnen:
The current survey ends, therefore we remove it from being triggered on the user side. The code remains (commented out) in place for later use and/or when we refactor the survey mechanism to work via the SonarLint website.


nquinquenel (approved):
LGTM but it really needs an automatic way of closing itself IMO, it will be displayed forever for users using the previous versions

PR #563, opened 2023-11-07
Use SLCORE through RPC

henryju:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/SonarLintEclipseHeadlessRpcClient.java
```diff
@@ -0,0 +1,185 @@
...
+    return false;
+  }
+
+  @Override
+  public void didReceiveServerHotspotEvent(DidReceiveServerHotspotEvent params) {
+    // No Hotspots in SLE
+  }
+
+  @Override
+  public void didUpdatePlugins(String connectionId) {
```
thahnen, line 145: Isn't it something similar to what we had prior in the `NodeJsManager` implementation?
> SonarLintCorePlugin.getConnectionManager().getConnections().forEach(f -> f.stop());

henryju: Is it a question for me? :)

thahnen: It was more for myself for later, but maybe you have an answer already 😄

henryju: Yes, I think stopping all the connected engines is the way to go.

PR #577, opened 2023-11-28
SLE-752: Deprecate username/password authentication

thahnen:
Users should only use tokens when setting up / changing connections in the future. Until then we provide them with information that the method is going to be deprecated in the future!


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/wizard/connection/UsernamePasswordWizardPage.java
```diff
@@ -72,6 +82,23 @@ protected void doCreateControl(Composite container) {
...
+    
+    var labelIcon = new Label(deprecationContainer, SWT.NULL);
+    labelIcon.setImage(SonarLintImages.IMG_SEVERITY_BLOCKER);
+    
+    var labelDeprecation = new Label(deprecationContainer, SWT.NULL);
+    labelDeprecation.setText(DEPRECATION_MESSAGE);
+    labelDeprecation.setLayoutData(new GridData(GridData.FILL_HORIZONTAL));
+    labelDeprecation.setFont(JFaceResources.getFontRegistry().getBold(JFaceResources.DEFAULT_FONT));
+    
+    labelIcon = new Label(deprecationContainer, SWT.NULL);
```
henryju, line 99: Maybe use different variables to clear confusion? `labelLeft` and `labelRight` or something similar.

henryju (approved):

PR #582, opened 2023-12-05
NO-JIRA: Log information about incomplete markers

thahnen:
This is based on the issue mentioned in this [Community Forum Thread](https://community.sonarsource.com/t/errors-and-seemingly-successful-analysis/105232). It is nice that we log the stack trace and the text range but we don't log the file/resource and the circumstances it is happening on - making it hard to debug such cases where it is not clear if the issue is coming from SonarLint or one of the language analyzers.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java
@@ -469,7 +469,7 @@ public class SonarLintMarkerUpdater {
         marker.setAttribute(IMarker.CHAR_START, flowPosition.getOffset());
         marker.setAttribute(IMarker.CHAR_END, flowPosition.getOffset() + flowPosition.getLength());
       } else {
-        SonarLintLogger.get().debug("Position cannot be set for flow on '" + file.getProjectRelativePath());
+        SonarLintLogger.get().debug("Position cannot be set for flow on '" + file.getProjectRelativePath() + "'");
       }
       return marker;
     }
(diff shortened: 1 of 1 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java
```diff
@@ -406,6 +409,8 @@ private static Optional<IMarker> createMarkerForTextRange(IDocument document, IR
         if (position != null) {
           marker.setAttribute(IMarker.CHAR_START, position.getOffset());
           marker.setAttribute(IMarker.CHAR_END, position.getOffset() + position.getLength());
+        } else {
+          SonarLintLogger.get().debug("Position cannot be set on resource '" + resource.getFullPath());
```
nquinquenel, line 413: Nitpick: 
```suggestion
          SonarLintLogger.get().debug("Position cannot be set on resource '" + resource.getFullPath() + "'");
```

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java
```diff
@@ -463,6 +468,8 @@ private static IMarker createMarkerIfCodeMatches(ISonarLintFile file, ServerIssu
       if (flowPosition != null) {
         marker.setAttribute(IMarker.CHAR_START, flowPosition.getOffset());
         marker.setAttribute(IMarker.CHAR_END, flowPosition.getOffset() + flowPosition.getLength());
+      } else {
+        SonarLintLogger.get().debug("Position cannot be set for flow on '" + file.getProjectRelativePath());
```
nquinquenel, line 472:
```suggestion
        SonarLintLogger.get().debug("Position cannot be set for flow on '" + file.getProjectRelativePath() + "'");
```

PR #583, opened 2023-12-05
SLE-784: Don't fail on incorrect classpath

thahnen:
This is based on the issue mentioned in this [Community Forum Thread](https://community.sonarsource.com/t/nullpointerexception-org-eclipse-core-resources-iresource-getlocation-is-null/105303). When a Java project classpath file contains invalid library entries we should log them but not fail the analysis.


damien-urruty-sonarsource (approved):
LGTM, but I would create a ticket at least for the change to be mentioned in the release notes (even if small)

PR #585, opened 2023-12-11
SLE-773: Update protobuf dependency

thahnen:
.

### DIFF
--- a/target-platforms/commons.target
+++ b/target-platforms/commons.target
@@ -17,7 +17,7 @@
 			  <dependency>
 				  <groupId>org.sonarsource.sonarlint.core</groupId>
 				  <artifactId>sonarlint-core-osgi</artifactId>
-				  <version>9.5.0.76302</version>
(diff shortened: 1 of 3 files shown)


target-platforms/commons.target
```diff
@@ -17,7 +17,7 @@
 			  <dependency>
 				  <groupId>org.sonarsource.sonarlint.core</groupId>
 				  <artifactId>sonarlint-core-osgi</artifactId>
-				  <version>9.5.0.76302</version>
+				  <version>9.6.0.76539</version>
```
henryju, line 20: Do we still have all languages enabled by default in SLE? Because a new language was added in 9.6, and I don't think we want it to be activated in SLE.

henryju (approved):
A question esle LGTM!

target-platforms/commons.target
thahnen: No, we enable them manually. When a new language is added we manually have to add it to SLE too.

PR #592, opened 2023-12-21
SLE-782: Fragments for Sloop

thahnen:
We are creating Sloop artifacts based on the JRE for macOS (x64/aarch64), Linux (x64/aarch64) and Windows (x64). For everything else we have a fallback bundle to provide their own JRE.


henryju (changes requested):
The noarch fragment seems to be installed in all cases, so it means we have twice the size of sloop downloaded for most cases (the noarch and the platform-specific). Ideally we should have the noarch only installed when no other fragment is suitable.

henryju:
image

thahnen:
@henryju the issue was fixed with the `Eclipse-PlatformFilter` now set as well in the MANIFEST.MF of the no-arch fragment

thahnen:
This PR is fucked up by rebasing -.-

PR #593, opened 2024-01-03
SLE-791: Automated connected mode setup

thahnen:
## Summary

When running "Open in IDE" with SonarQube 10.4+, we can automatically set up connected mode. This helps the user to have a minimal solution for configuration.

## Testing

Integration tests were added to the already existing "Open in IDE" tests, and the whole test suite was overhauled to only create a project once on SQ and to run Maven only once in order to save [precious](https://encrypted-tbn0.gstatic.com/images?q=tbn:ANd9GcQRs65wa-OyfSBvrvzmqkQfUR4fz44neeZniGmOIXy07A&s) build time.

### DIFF
--- a/its/pom.xml
+++ b/its/pom.xml
@@ -164,8 +164,8 @@
           <appArgLine>-eclipse.keyring target/keyring -eclipse.password secure-storage-password.txt -pluginCustomization
             ${project.basedir}/plugin_customization.ini</appArgLine>
           
-          <!-- Kill test JVM if tests take more than 45 minutes (2700 seconds) to finish -->
-          <forkedProcessTimeoutInSeconds>2700</forkedProcessTimeoutInSeconds>
+          <!-- Kill test JVM if tests take more than 60 minutes (3600 seconds) to finish -->
+          <forkedProcessTimeoutInSeconds>3600</forkedProcessTimeoutInSeconds>
           
           <redirectTestOutputToFile>true</redirectTestOutputToFile>
           <systemProperties>
--- a/its/src/org/sonarlint/eclipse/its/OpenInIdeTest.java
+++ b/its/src/org/sonarlint/eclipse/its/OpenInIdeTest.java
@@ -29,16 +29,19 @@ import java.net.http.HttpClient;
 import java.net.http.HttpRequest;
 import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 import org.eclipse.reddeer.common.wait.TimePeriod;
 import org.eclipse.reddeer.common.wait.WaitUntil;
 import org.eclipse.reddeer.core.condition.WidgetIsFound;
 import org.eclipse.reddeer.core.exception.CoreLayerException;
 import org.eclipse.reddeer.core.matcher.WithTextMatcher;
+import org.eclipse.reddeer.eclipse.jdt.ui.wizards.NewClassCreationWizard;
+import org.eclipse.reddeer.eclipse.jdt.ui.wizards.NewClassWizardPage;
 import org.eclipse.reddeer.eclipse.ui.perspectives.JavaPerspective;
 import org.eclipse.reddeer.swt.impl.shell.DefaultShell;
 import org.eclipse.reddeer.workbench.impl.editor.TextEditor;
 import org.eclipse.swt.widgets.Label;
-import org.junit.After;
+import org.junit.AfterClass;
 import org.junit.Assume;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
@@ -59,6 +62,7 @@ import org.sonarqube.ws.Issues.Issue;
 import org.sonarqube.ws.WsBranches.Branch;
 import org.sonarqube.ws.client.issue.SearchWsRequest;
 import org.sonarqube.ws.client.project.DeleteRequest;
+import org.sonarqube.ws.client.usertoken.GenerateWsRequest;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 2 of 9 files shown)


its/pom.xml
```diff
@@ -164,8 +164,8 @@
           <appArgLine>-eclipse.keyring target/keyring -eclipse.password secure-storage-password.txt -pluginCustomization
             ${project.basedir}/plugin_customization.ini</appArgLine>
           
-          <!-- Kill test JVM if tests take more than 45 minutes (2700 seconds) to finish -->
-          <forkedProcessTimeoutInSeconds>2700</forkedProcessTimeoutInSeconds>
+          <!-- Kill test JVM if tests take more than 60 minutes (3600 seconds) to finish -->
+          <forkedProcessTimeoutInSeconds>3600</forkedProcessTimeoutInSeconds>
```
thahnen, line 168: We hit the mark too often, that is why I increased it. I've spoken to RE and we ended up with [THIS](https://sonarsource.atlassian.net/browse/BUILD-4406) follow-up ticket, that I'm going to tackle after this one is merged - build times should decrease.

its/src/org/sonarlint/eclipse/its/OpenInIdeTest.java
```diff
@@ -196,25 +261,27 @@ public void test_open_in_ide_when_project_already_bound() throws IOException, In
...
     new WaitUntil(new OpenInIdeDialogOpened(), TimePeriod.DEFAULT);
     new OpenInIdeDialog().ok();
 
-    // 6) change preferences regarding new code / issue filter + change file content
+    // 5) change preferences regarding new code / issue filter + change file content
     setNewCodePreference(IssuePeriod.NEW_CODE);
-    openFileAndWaitForAnalysisCompletion(rootProject.getResource("src/main/java", "taint", "SysOut.java"));
+    
+    rootProject.getResource("src/main/java", "taint", "SysOut.java").delete();
+    createClass("taint", "SysOut.java");
```
thahnen, line 284: FYI: Doing it this way as the other option was too flaky and broke easily.

its/pom.xml
henryju: Maybe we should rely only on Cirrus timeout, and not on this Maven feature at all? This look like a remain of the past, AFAIK no other project is using surefire timeout.

its/src/org/sonarlint/eclipse/its/OpenInIdeTest.java
```diff
@@ -80,7 +84,7 @@ public class OpenInIdeTest extends AbstractSonarQubeConnectedModeTest {
     .keepBundledPlugins()
     .setEdition(Edition.DEVELOPER)
     .activateLicense()
-    .setSonarVersion(System.getProperty("sonar.runtimeVersion", "LATEST_RELEASE"))
+    .setSonarVersion(System.getProperty("sonar.runtimeVersion", "DEV"))
```
henryju, line 87: Is it a temporary fix? Do we really want to test against SQ DEV in the long run? If not, maybe add a comment explaining why we had to use DEV.

its/src/org/sonarlint/eclipse/its/OpenInIdeTest.java
henryju: Maybe add a comment then

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/assist/AssistCreatingAutomaticConnectionJob.java
```diff
@@ -0,0 +1,43 @@
...
+  public AssistCreatingAutomaticConnectionJob(String serverUrl, String tokenValue) {
+    super("Assist automatic creation of connected mode", serverUrl, true);
+    this.tokenValue = tokenValue;
+  }
+
+  @Override
+  @Nullable
+  protected IConnectedEngineFacade createConnection(ServerConnectionModel model) {
+    var connection = SonarLintCorePlugin.getServersManager().create(model.getConnectionId(), model.getServerUrl(),
+      model.getOrganization(), tokenValue, model.getPassword(), model.getNotificationsDisabled());
```
henryju, line 39:
```suggestion
      model.getOrganization(), tokenValue, null, model.getNotificationsDisabled());
```

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/assist/AssistCreatingAutomaticConnectionJob.java
```diff
@@ -0,0 +1,43 @@
...
+    SonarLintCorePlugin.getServersManager().addServer(connection, tokenValue, model.getPassword());
```
henryju, line 40:
```suggestion
    SonarLintCorePlugin.getServersManager().addServer(connection, tokenValue, null);
```

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/backend/SonarLintEclipseClient.java
```diff
@@ -161,20 +165,36 @@ public void showHotspot(ShowHotspotParams params) {
...
         if (job.getResult().isOK()) {
           SonarLintLogger.get().debug("Successfully created connection '" + job.getConnectionId() + "'");
           return new AssistCreatingConnectionResponse(job.getConnectionId());
-        } else if (job.getResult().getCode() == IStatus.CANCEL) {
+        } else if (job.getResult().matches(IStatus.CANCEL)) {
+          if (job instanceof AssistCreatingAutomaticConnectionJob) {
+            SonarLintBackendService.get()
+              .getBackend()
+              .getUserTokenService()
+              .revokeToken(new RevokeTokenParams(baseUrl, params.getTokenName(), params.getTokenValue()));
```
henryju, line 190: This is a notification, right? If not, you should call `join()` on the CompletableFuture.

henryju (approved):
A few minor comments, but overall LGTM! Well done!

its/src/org/sonarlint/eclipse/its/OpenInIdeTest.java
thahnen: Done

its/src/org/sonarlint/eclipse/its/OpenInIdeTest.java
thahnen: Missed that

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/assist/AssistCreatingAutomaticConnectionJob.java
thahnen: Done

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/backend/SonarLintEclipseClient.java
thahnen: Yes, it is. There is no return value and in the end, the likelihood that it fails is near then zero so it shouldn't block the thread but it can take some time due to the network connection.

It can only _fail_ on three occasions:
- network down (unlikely as we just had contact from the SQ to invoke the "Open in IDE", maybe someone let the confirmation window open and turned off the network connection - highly unlikely)
- the server is down (same as above)
- unauthorized (token was just granted, someone on SQ must have removed it just seconds ago)

henryju: If it is a notification, it will be the responsibility of the core to handle the errors, so that's fine.

PR #596, opened 2024-01-23
SLE-408: Notification about analysis accuracy

thahnen:
When the automatic build of the workspace is disabled, some rules might not yield accurate results compared to when the automatic build of the workspace is enabled.

### DIFF
--- a/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
+++ b/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
@@ -50,6 +50,7 @@ import org.eclipse.reddeer.workbench.impl.editor.Marker;
 import org.eclipse.reddeer.workbench.impl.editor.TextEditor;
 import org.eclipse.reddeer.workbench.ui.dialogs.WorkbenchPreferenceDialog;
 import org.hamcrest.core.StringContains;
+import org.junit.After;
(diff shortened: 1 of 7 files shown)


its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
```diff
@@ -74,6 +76,56 @@ public class StandaloneAnalysisTest extends AbstractSonarLintTest {
...
+    if ("oldest".equals(System.getProperty("target.platform"))) {
+      var preferences = GeneralWorkspaceBuildPreferences.open();
+      preferences.enableAutomaticBuild();
+      preferences.ok();
+    }
+  }
+  
+  @Test
+  public void analyze_automatic_workspace_build_disabled() {
+    // INFO: We only want to run it on one axis and the "oldest" ITs take the shortest!
```
henryju, line 93: On the long run, it might be cleaner to use JUnit categories to partition tests.

henryju (approved):
Sounds like a nice addition, well done!

its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
thahnen: I haven't worked with them yet, but it is a valid point!

PR #599, opened 2024-01-25
SLE-792: Users can detect their own secret patterns

thahnen:
...


thahnen:
This got reworked due to the constrains on `sonarlint-core`, see [THIS](https://github.com/SonarSource/sonarlint-eclipse/pull/602) pr.

PR #600, opened 2024-01-26
Rename Server* -> Connection*

henryju:
My PR with the migration to RPC is gigantic, with a mix of small refactoring I made along the way, but that are not strictly speaking necessary for RPC migration.
So I am extracting the changes in a separate PR, hopefully it will make review simpler.


thahnen (approved):
LGTM! I assume most of the whitespace stuff is coming from changes I did in most of the files, but I don't understand why 😮 
I'm using the stuff from [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset) on every workspace, maybe it is outdated for the whitespace? I was under the assumption that saving the file would clean this and stuff like the unused imports.

henryju:
> LGTM! I assume most of the whitespace stuff is coming from changes I did in most of the files, but I don't understand why 😮 I'm using the stuff from [SonarSource/sonar-developer-toolset](https://github.com/SonarSource/sonar-developer-toolset) on every workspace, maybe it is outdated for the whitespace? I was under the assumption that saving the file would clean this and stuff like the unused imports.

Do you have "Save Action" enabled?

image

PR #604, opened 2024-02-01
SLE-799: Overhaul rules configuration messaging

thahnen:
# Summary

The existing message plugged into the saving of the local rules configuration was overhauled in phrasing. Additionally, a switch was added for users configuring it despite all projects already bound.

# Testing

It was manually tested because Reddeer **always** expects [to come back to the main window after the preferences are closed](https://github.com/eclipse/reddeer/blob/master/plugins/org.eclipse.reddeer.jface/src/org/eclipse/reddeer/jface/preference/PreferenceDialog.java#L118-L120). It currently does not seem to support new shells popping up before the focus is back at the main window.

[THIS](https://github.com/eclipse/reddeer/issues/2227) issue was created. I've read in some threads that Reddeer is not maintained anymore but I didn't find anything official. Let's hope it is still maintained.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/ProjectsProviderUtils.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/ProjectsProviderUtils.java
@@ -51,7 +51,7 @@ public class ProjectsProviderUtils {
     var allProjects = allProjects();
     var numberOfAllProjects = allProjects.size();
     var boundProjects = allProjects.stream()
-      .filter(prj -> SonarLintCorePlugin.getServersManager().resolveBinding(prj).isPresent())
+      .filter(prj -> SonarLintCorePlugin.getConnectionManager().resolveBinding(prj).isPresent())
       .collect(Collectors.toSet());
     var numberOfBoundProjects = boundProjects.size();
     return numberOfAllProjects == 0 ? 0 : (numberOfBoundProjects / numberOfAllProjects);
@@ -59,6 +59,6 @@ public class ProjectsProviderUtils {
 
   public static Set<String> allConfigurationScopeIds() {
     return allProjects().stream().map(ConfigScopeSynchronizer::getConfigScopeId)
-    .collect(Collectors.toSet());
+      .collect(Collectors.toSet());
   }
 }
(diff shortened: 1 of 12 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/ProjectsProviderUtils.java
```diff
@@ -40,8 +41,24 @@ public static Collection<ISonarLintProject> allProjects() {
       .collect(Collectors.toSet());
   }
 
+  /**
+   *  Useful when we want SonarLint to behave differently when
+   *  - no project bound (ret = 0)
+   *  - all projects bound (ret = 1)
+   *  - at least one project bound (0 < ret < 1)
+   */
+  public static float boundToAllProjectsRatio() {
```
henryju, line 50: Do you really make any use of this float? Why not return an enum:
```
enum WorkspaceProjectsBindingRatio {
    NONE_BOUND,
    SOME_BOUND,
    ALL_BOUND
}
```
It will make calling code more readable IMO.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/job/RulesConfigurationPageSaveJob.java
```diff
@@ -0,0 +1,55 @@
...
+/** Job triggered when the rules configuration page changes are saved in order to not block the UI */
+public class RulesConfigurationPageSaveJob extends Job {
+  public RulesConfigurationPageSaveJob() {
+    super("Saving the rules configuration");
+  }
+
+  @Override
+  protected IStatus run(IProgressMonitor monitor) {
+    var boundToAllProjectsRatio = ProjectsProviderUtils.boundToAllProjectsRatio();
+    if (boundToAllProjectsRatio == 1f) {
```
henryju, line 40: You see, here, I think you could save the comment next line by having code like:
```suggestion
    if (projectsBindingRatio == ALL_BOUND) {
```

henryju (approved):
I would avoid returning a float in place of a 3-state result, but else LGTM.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/job/RulesConfigurationPageSaveJob.java
thahnen: Done

PR #607, opened 2024-02-02
BUILD-4406: Consistent pipeline times

thahnen:
This is to be used as an intermediate solution. So we can continue with the optimization process on our side, we want to have build times that are not jumping up and down because Cirrus is choosing a build node type at random.


nquinquenel (approved):
💵

PR #609, opened 2024-02-06
SLE-802: Re-evaluated the preference persistence

thahnen:
Some preferences persisted on the wrong level and were therefore interfering (but not breaking) some UI / UX stuff (e.g. notifications about Taint Vulnerabilities introduced).
This was re-evaluated and changed.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintGlobalConfiguration.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintGlobalConfiguration.java
@@ -37,6 +37,7 @@ import org.eclipse.core.resources.IMarker;
 import org.eclipse.core.runtime.Platform;
 import org.eclipse.core.runtime.preferences.ConfigurationScope;
 import org.eclipse.core.runtime.preferences.IEclipsePreferences;
+import org.eclipse.core.runtime.preferences.InstanceScope;
 import org.eclipse.jdt.annotation.Nullable;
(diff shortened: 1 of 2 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintGlobalConfiguration.java
```diff
@@ -266,7 +271,7 @@ private static Set<RuleConfig> deserializeRulesJson(String json) {
 
   public static void saveRulesConfig(Collection<RuleConfig> rules) {
     var json = serializeRulesJson(rules);
-    setPreferenceString(PREF_RULES_CONFIG, json);
+    setPreferenceString(getApplicationLevelPreferenceNode(), PREF_RULES_CONFIG, json);
```
henryju, line 274: Isn't it a breaking change? People having previously customized their rule config (stored in the workspace scope) will lose their config.
You should maybe implement a migration when reading (try to read from the old location, and if present, copy the settings to the new location).

henryju (approved):
I am a bit worried about the effects of the changes on the users, but on a technical PoV LGTM.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintGlobalConfiguration.java
henryju: Ah, you are right, sorry. Then LGTM!

henryju (approved):

PR #610, opened 2024-02-09
SLE-782: Sloop fragments based on SLE 9.4

thahnen:
Because the former [PR](https://github.com/SonarSource/sonarlint-eclipse/pull/592) was messed up due to rebasing a lot of changes.

### DIFF
--- a/org.sonarlint.eclipse.sloop.no-arch/META-INF/MANIFEST.MF
+++ b/org.sonarlint.eclipse.sloop.no-arch/META-INF/MANIFEST.MF
@@ -1,0 +1,11 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: %bundle_name
+Bundle-SymbolicName: org.sonarlint.eclipse.sloop.no-arch;singleton:=true
+Bundle-Version: 9.4.0.qualifier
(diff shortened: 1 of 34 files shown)


org.sonarlint.eclipse.sloop.no-arch/META-INF/MANIFEST.MF
```diff
@@ -0,0 +1,11 @@
+Manifest-Version: 1.0
+Bundle-ManifestVersion: 2
+Bundle-Name: %bundle_name
+Bundle-SymbolicName: org.sonarlint.eclipse.sloop.no-arch;singleton:=true
+Bundle-Version: 9.4.0.qualifier
+Eclipse-PlatformFilter: (& (osgi.os=win32) (osgi.arch=aarch64))
```
nquinquenel, line 6: is this valid for no arch?

thahnen: Yes. There is currently no option to have a filter that says "everything else." I opened a [discussion](https://github.com/eclipse-pde/eclipse.pde/discussions/1026) at Eclipse PDE some time ago.

Currently, we only allow macOS, Linux, and Windows with the architectures x86_64 and aarch64 (ARM). The only one we have not with a Java distribution is Windows on ARM, because there basically is none.

nquinquenel (approved):
Looks good

PR #613, opened 2024-02-15
(WIP) SLE-781: SonarLint Core out of Process

thahnen:
This is a follow-up to #563


its/src/org/sonarlint/eclipse/its/reddeer/conditions/AnalysisReady.java
thahnen, line 1: Git is a bit lost, this file wasn't moved. This was created and the other one is completely gone.

damien-urruty-sonarsource (approved):
Let's merge this MF!

PR #614, opened 2024-02-20
Update to Tycho 4.0.6 and provision tests with p2

henryju:
In order to have more "realistic" behavior of tests, we should switch to `p2Installed` mode for surefire.

This will allow to execute p2 touchpoints like a real installation.

See discussion: https://github.com/eclipse-tycho/tycho/discussions/3440

### DIFF
--- a/its/its-ibuilds.product
+++ b/its/its-ibuilds.product
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<?pde version="3.5"?>
+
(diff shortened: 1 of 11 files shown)


henryju:
@thahnen thanks to the recent release of Tycho, I was able to use p2Installed mode for running tests and ITs. This should allow to remove the hack for fixing permissions on the sloop branch.

thahnen:
Hi @henryju,

thank you for working on this. I will have a look once I'm done testing the Sloop stuff.
Afterwards we can then merge it 👍

thahnen (approved):
LGTM, even though I don't fully understand how it works yet, it looks good and everything seems to work!
Please rebase once and then merge it 😄

PR #625, opened 2024-03-06
SLE-807: CycloneDx only on necessary build steps

thahnen:
This speeds up the local builds and also some of the CI tasks


henryju (approved):
LGTM

PR #626, opened 2024-03-06
BUILD-3457 publish dogfood site on binaries

julien-carsique-sonarsource:
Instead of promoting the site artifact to sonarlint-eclipse-dogfood which is not publicly accessible, publish the site to https://binaries.sonarsource.com/?prefix=SonarLint-for-Eclipse/dogfood/

Manually tested with staging, see https://binaries-staging.sonarsource.com/?prefix=SonarLint-for-Eclipse/dogfood/

### DIFF
--- a/.cirrus.default.yml
+++ b/.cirrus.default.yml
@@ -96,6 +96,9 @@ build_task:
     base64 --decode "${SM_CLIENT_CERT_FILE}.b64" > "${SM_CLIENT_CERT_FILE}"
     source cirrus-env BUILD-PRIVATE
     .cirrus/regular_mvn_build_deploy_analyze -Dmaven.test.skip=true -Dsonar.skip=true -Dcyclonedx.skip=false
+  site_artifacts:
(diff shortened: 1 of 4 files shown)


thahnen (changes requested):
Hi @julien-carsique-sonarsource,

thank you for making this possible!
Two issues I have with this tho: What are the `.envrc` files for? Are these coming from your local dev environment? (looks like it because one is referencing the staging binaries URL)
If so, please get rid of them before the merge.

The second thing: Could you please move everything to the `.cirrus` directory? That would be great as it is only to be triggered from there 😄

julien-carsique-sonarsource:
@thahnen The .envrc files are used for local development. They would be useful for other people if working on the same topics, but I can remove them. https://direnv.net/

I'm moving the files from scripts/ to .cirrus/

thahnen (approved):
LGTM!

If the build fails, this is due to #627. I'm sorry for that, I'll merge it as quickly as possible and then take care of rebasing your branch onto master.

PR #630, opened 2024-03-08
SLE-818: Sign ProtoBuf OSGi bundle

thahnen:
The logic for signing ProtoBuf was incorrect and only signed the source OSGi bundle.


henryju (approved):
Good catch!

PR #631, opened 2024-03-12
SLE-822: Enable users to set JRE for Sloop

thahnen:
This will enable power users to configure SonarLint to use a specific Java 17+ JRE to run SonarLint out of process.

This is by default configured via the workspace settings but can be configured on the application level as well (not from the IDE).

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/documentation/SonarLintDocumentation.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/documentation/SonarLintDocumentation.java
@@ -38,6 +38,7 @@ public class SonarLintDocumentation {
   public static final String RULES = BASE_DOCS_URL + "/using-sonarlint/rules/";
   public static final String RULES_SELECTION = RULES + "#rule-selection";
   public static final String TROUBLESHOOTING_LINK = BASE_DOCS_URL + "/troubleshooting/";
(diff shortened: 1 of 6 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/documentation/SonarLintDocumentation.java
```diff
@@ -38,6 +38,7 @@ public class SonarLintDocumentation {
   public static final String RULES = BASE_DOCS_URL + "/using-sonarlint/rules/";
   public static final String RULES_SELECTION = RULES + "#rule-selection";
   public static final String TROUBLESHOOTING_LINK = BASE_DOCS_URL + "/troubleshooting/";
+  public static final String ADVANCED_CONFIGURATION = BASE_DOCS_URL + "team-features/advanced-configuration/";
```
nquinquenel, line 41: I think it's missing a `/` at the beginning

nquinquenel (approved):
LGTM

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/documentation/SonarLintDocumentation.java
thahnen: True, I fixed that

PR #640, opened 2024-04-03
DO NOT MERGE: Use LSP4J 0.15.0 instead of 0.21.0

thahnen:
TODO: Follow up with information


thahnen:
WIP: This was done just for testing on feasibility.

PR #641, opened 2024-04-17
NO-JIRA: Fix master build

thahnen:
This should resolve the [Maven lifecycle issue](https://cirrus-ci.com/task/6255246911995904) and should fail faster in the future.

The issue was that the SLCORE dependencies weren't available on Repox anymore, and additionally, the output of Maven / Tycho indicated otherwise and misled to another Eclipse-related issue based on the dependency requirement of SLE on SLCORE artifacts.

The implication of this change is that every update of SLCORE in SLE requires the manifest files to be changed. The upside of this is that user installations with multiple versions of the dependencies will only use the ones with the correct version.


org.sonarlint.eclipse.core/META-INF/MANIFEST.MF
```diff
@@ -48,9 +48,9 @@ Require-Bundle: org.eclipse.equinox.security,
  org.eclipse.text,
  org.eclipse.lsp4j.jsonrpc,
  com.google.protobuf;bundle-version="3.25.1",
- org.sonarsource.sonarlint.core.sonarlint-rpc-protocol;bundle-version="10.0.0",
- org.sonarsource.sonarlint.core.sonarlint-rpc-java-client;bundle-version="10.0.0",
- org.sonarsource.sonarlint.core.sonarlint-java-client-utils;bundle-version="10.0.0",
- org.sonarsource.sonarlint.core.sonarlint-java-client-legacy;bundle-version="10.0.0"
+ org.sonarsource.sonarlint.core.sonarlint-rpc-protocol;bundle-version="[10.1.0,10.2.0)",
```
thahnen, line 51: FYI: This will only allow SLCORE packages with the correct version (currently SLE 10.1 is also targeting SLCORE 10.1). If I try to update to an unsupported version and run Maven locally, this will fail it fast.

pom.xml
```diff
@@ -71,7 +71,7 @@
     <jarsigner.version>3.0.0</jarsigner.version>
 
     <!-- Sloop embedded CLI version for fragment projects -->
-    <sloop.version>10.1.0.77273</sloop.version>
+    <sloop.version>10.1.0.77633</sloop.version>
```
jblievremont, line 74: This version number is duplicated several times. Would it be possible e.g. to extract it to a property file and refer to it in the `.target` file?

Otherwise, I would add a comment here to remind the next author to update the other files (both `MANIFEST.MF`s + `commons.target`).

jblievremont (approved):
LGTM in terms of bringing `master` back to 🟢!

I'm not sure if my other comment makes sense within the limits of the current build system, let's keep this in mind in case the problem comes back.

PR #642, opened 2024-04-17
SLE-819: Add support for JCL

thahnen:
Because the other [PR](https://github.com/SonarSource/sonarlint-eclipse/pull/637) couldn't be merged anymore from within GitHub.


jblievremont (approved):
LGTM 🚢 

As a bonus, would it be possible to add an IT with a small JCL sample?

PR #643, opened 2024-04-18
SLE-843 Use analysis execution from the backend

damien-urruty-sonarsource:


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/SonarLintEclipseHeadlessRpcClient.java
```diff
@@ -59,6 +67,16 @@
...
  */
 public abstract class SonarLintEclipseHeadlessRpcClient implements SonarLintRpcClientDelegate {
+  
+  @Override
+  public Path getBaseDir(String configurationScopeId) throws ConfigScopeNotFoundException {
+    var project = resolveProject(configurationScopeId);
+    var projectLocation = project.getResource().getLocation();
+    // In some unfrequent cases the project may be virtual and don't have physical location
+    // so fallback to use analysis work dir that was created before
+    var analysisWorkDir = project.getWorkingDir().resolve("sonarlint");
```
thahnen, line 77: This doesn't make any sense to me. Where is the magic `sonarlint` folder coming from? I've never encountered that and I also didn't find any clue where this folder would have been created.

damien-urruty-sonarsource: I just reused the same piece of code that we were using before, see https://github.com/SonarSource/sonarlint-eclipse/blob/10.0.1.81733/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AbstractAnalyzeProjectJob.java#L180

damien-urruty-sonarsource: And the continuation here: https://github.com/SonarSource/sonarlint-eclipse/blob/10.0.1.81733/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AbstractAnalyzeProjectJob.java#L259

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AnalyzeProjectJob.java
```diff
@@ -474,13 +439,37 @@ public Map<String, String> getAnalysisProperties() {
...
+      throw new CanceledException();
+    } catch (ExecutionException e) {
+      throw new IllegalStateException(e);
+    } finally {
+      RunningAnalysesTracker.get().finish(analysisState);
+    }
   }
 
-  protected abstract AnalysisResults runAnalysis(AnalysisConfiguration analysisConfig, SonarLintIssueListener issueListener, IProgressMonitor monitor);
+  private static <T> T waitForFuture(IProgressMonitor monitor, CompletableFuture<T> future) throws InterruptedException, ExecutionException {
```
thahnen, line 444: This implementation is not needed, we already have a [JobUtils.waitForFuture](https://github.com/SonarSource/sonarlint-eclipse/blob/10.0.1.81733/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/JobUtils.java#L53-L64), please re-use this one.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java
```diff
@@ -366,41 +368,41 @@ private static void createQuickFixMarkersForLocalIssues(IDocument document, ISon
...
-        SonarLintLogger.get().debug("Document has changed since quick fix was contributed: " + engineQuickFix.message());
+  private static void createQuickFix(IDocument document, ISonarLintIssuable issuable, List<MarkerQuickFix> qfs, QuickFixDto rpcQuickFix) {
+    var qf = new MarkerQuickFix(rpcQuickFix.message());
+    for (var edits : rpcQuickFix.fileEdits()) {
+      var fileWithEdit = SonarLintUtils.findFileFromUri(edits.target());
+      if (!issuable.equals(fileWithEdit)) {
+        SonarLintLogger.get().debug("Quick fix on multiple files is not supported yet: " + rpcQuickFix.message());
         return;
       }
+      // TODO
```
thahnen, line 385: Get rid of this, please 😄

thahnen (approved):
Looks good to me!

The `SonarLintUtils#findFileFromUri` I will have to check with IBM IDz for COBOL, other than that everything looks either fine or is mentioned in a newly created ticket.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/AnalyzeProjectJob.java
damien-urruty-sonarsource: The `JobUtils.waitForFuture` behavior is a bit different than the method I created here, because I wanted to explicitly throw a `CancelationException` in the context of a `Job`. But I moved the method to `JobUtils` so it can be reused (I renamed it to `waitForFutureInJob`), and at some point we might need to harmonize how we handle those cases in SLE.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java
damien-urruty-sonarsource: I removed the dead code, I just left a question instead

PR #647, opened 2024-04-23
Analyzer update

thahnen:
This includes updates of the following analyzers:
- Java
- JS/TS
- PHP
- Python
- HTML
- SonarText


sophio-japharidze-sonarsource (approved):
LGTM! Just to make sure, should no reference to Node runtime requirement be updated?

PR #652, opened 2024-04-26
SLE-854: Fix JDT with COBOL

thahnen:
When having the Java Development Tools (JDT) installed but working with COBOL files with non-Java projects, we have to make sure JDT does not try to see it as a 'Java-like' file due to the extension (e.g., `cbl`). Otherwise, it would try to exclude it from the analysis because it is not in the Java classpath and, therefore, not compiled.

JDT itself tries to find 'Java-like' file extensions by the standard Java file suffixes and from other ones it seems. I've created an [issue](https://github.com/eclipse-jdt/eclipse.jdt.core/issues/2389) with Eclipse JDT just in case this is actually a bug.

### DIFF
--- a/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
+++ b/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
@@ -66,6 +66,7 @@ import org.sonarlint.eclipse.its.reddeer.preferences.SonarLintProperties;
 import org.sonarlint.eclipse.its.reddeer.views.OnTheFlyView;
 import org.sonarlint.eclipse.its.reddeer.views.PydevPackageExplorer;
 import org.sonarlint.eclipse.its.reddeer.views.ReportView;
+import org.sonarlint.eclipse.its.reddeer.views.SonarLintConsole;
 import org.sonarlint.eclipse.its.reddeer.views.SonarLintIssueMarker;
 
 import static org.assertj.core.api.Assertions.assertThat;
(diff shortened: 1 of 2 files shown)


its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
```diff
@@ -128,6 +129,19 @@ public void analyze_automatic_workspace_build_disabled() {
...
+  @Test
+  public void test_jdt_java_like_extension_COBOL() {
+    new JavaPerspective().open();
+    var rootProject = importExistingProjectIntoWorkspace("connected", "connected");
+
+    var cobolFile = rootProject.getResource("Test.cbl");
+    openFileAndWaitForAnalysisCompletion(cobolFile);
+
+    assertThat(new SonarLintConsole().getConsoleView().getConsoleText())
+      .doesNotContain("File 'Test.cbl' excluded by 'JavaProjectConfiguratorExtension'");
```
jblievremont, line 142: Is there a way to make sure that the file was actually analyzed, at least by the secrets analyzer?

jblievremont (approved):
LGTM from a code point of view, I would love to see a "positive" assertion in the test.

its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
thahnen: Done, the test is changed to test for both.

PR #656, opened 2024-05-03
SLE-793: Dependency simplification

thahnen:
## Summay

In order to fix issues of leaking packages that created conflicts with other Eclipse plug-ins/bundles, the new dependencies on SLCORE are once again shaded/relocated into a single OSGi bundle. The main changes are done on SLCORE; on SLE, there is only relying on the relocated GSON.

The bundle `sonarlint-java-client-osgi` contains the other SLCORE modules as well as the dependencies *LSP4J*, *GSON*, and *ProtoBuf*. This mimics the pre-Sloop logic of having the `sonarlint-core-osgi` bundle.

## Testing

Integration tests running successfully, manually tested different Eclipse installation and plug-in combinations. Testing with COBOL-IDEs will be done on their side, already reached out.


nquinquenel (approved):
Looks much more cleaner with this single bundle!

PR #658, opened 2024-05-13
SLE-861: Less friction with the Java installation used by SonarLint

thahnen:
## Summary

When SonarLint for Eclipse is installed on a machine that does not allow the execution of unspecified Java installations (e.g. Windows Group Policy issues), the backend won't start. The user should configure a self-managed Java installation in that case via the SonarLint preference page.

This is hideous if the user already configured Eclipse (or the Eclipse-based IDE) to use a self-managed Java installation on the one hand, and on the other hand, changing the Java installation via the SonarLint preference page does not work because it relies on the SonarLint backend up and running.

As the Java installation bundled with SonarLint was intended in the first place for IDEs still running on Java 11, in the case of IDEs running on Java 17+, we don't "need" our Java runtime and can re-use the one configured/used by Eclipse. This also reduces the overhead for power users regarding configuration!

The SonarLint preference pages were changed as well to not assume that the SonarLint backend is up and running, at least for users to change the self-managed Java installation. There should be no UI error or exception anymore when navigating the preference pages. Only saving different combinations that rely on the SonarLint backend might not work, but that is expected!

For debugging this issue in the first place and later for making our lives easier, an abstraction that was introduced for `Adapters.adapt(...)` a while ago is now used everywhere where we try to adapt. This helps to trace potential issues even better. This was done specifically in a separate commit!

## Testing

In order to create a not-working SonarLint installation on Eclipse, the preference `/instance/org.sonarlint.eclipse.ui/java17Path` must be set to something random (e.g., no working Java installation home directory). After that, it can be tested that the preferences are set up correctly, and changing the Java 17 runtime to a self-managed installation is possible.

In the future, it might be possible that we create additional integration test projects where we test that running SonarLint with a self-managed as well as Eclipse-provided installation and no working installation at all is possible. **This is not in the scope of this ticket or hardening**!


nquinquenel (approved):
LGTM (did not test)

PR #659, opened 2024-05-14
Avoid resourceChanged computation for IncrementalProjectBuilder.CLEAN_BUILD in FileSystemSynchronizer

RoiSoleil:
As mentionned in the IncrementalProjectBuilder.CLEAN_BUILD documentation 'Resource deltas are not applicable for this kind of build.'
This is a performance optimisation. I did noticed that when i was doing a "Maven update" is was way longer when SonarLint is enabled. It was because the FileSystemSynchronizer #resourceChanged was doing a lot of computation.
It's the same issue than https://github.com/SonarSource/sonarlint-eclipse/pull/539

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/FileSystemSynchronizer.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/FileSystemSynchronizer.java
@@ -32,6 +32,7 @@ import org.eclipse.core.filesystem.EFS;
 import org.eclipse.core.resources.IResourceChangeEvent;
 import org.eclipse.core.resources.IResourceChangeListener;
 import org.eclipse.core.resources.IResourceDelta;
+import org.eclipse.core.resources.IncrementalProjectBuilder;
 import org.eclipse.core.runtime.Adapters;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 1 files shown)


thahnen:
Hi @RoiSoleil,

thank you again (!) for opening a PR to improve the performance of SonarLint for Eclipse. I will take a look tomorrow and get a pipeline running for this.

Best,
Tobias

RoiSoleil:
Thx for the quick answer :)

thahnen:
Hi @RoiSoleil,

I looked at the changes you made, and currently (for the context of running `Maven Update`), they don't have an impact. I've analyzed what the Update is actually triggering, and it's not an `IncrementalProjectBuilder.CLEAN_BUILD`, but the Maven integration of Eclipse (m2e) is rather providing events for "added" or "changed" files.

Even though the files were not added or changed (some say it is about the encoding, that is not correct) SonarLint itself works correctly and handles the files correctly in that case. It seems to be either m2e or Eclipse that doesn't respond correctly to the `Maven Update` request.

I've seen that the performance might be increased (but that is for all our `IResourceChangeListener`) and I will create a ticket for that.

I've tested with **Eclipse RCP 2024-03**, **SonarLint for Eclipse 10.1** (and our latest Dogfooding version) with the latest **M2E installation 2.6.0.20240220-1109**.

RoiSoleil:
You're right sorry for the inconvenience ... I did open a PR in M2E to avoid the encoding change to occur : https://github.com/eclipse-m2e/m2e-core/pull/1758

thahnen:
Thanks for the heads up; when there is a new (dogfooding) version of m2e, please give it a try or ping me. With the latest released version of SonarLint for Eclipse, there is more aggressive logging at least so changes on m2e should be visible by having less logs in that case.

PR #660, opened 2024-05-14
SLE-793: Use simplified SLCORE artifact

thahnen:
This uses the simplified artifact from https://github.com/SonarSource/sonarlint-core/pull/970. The simplification in the OSGI bundle should have no impact.


vojtech-suchy-sonarsource (approved):
Excellent

PR #661, opened 2024-05-15
SLE-851: Handle simple and complex timeout configuration

thahnen:
With that SonarLint can now handle simple (like "3" for "3 minutes") and complex (like "PT4H" for "4 hours") timeout configurations again.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/DurationUtils.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/DurationUtils.java
@@ -1,0 +1,55 @@
+/*
+ * SonarLint for Eclipse
+ * Copyright (C) 2015-2024 SonarSource SA
+ * sonarlint@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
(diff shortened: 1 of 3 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/DurationUtils.java
```diff
@@ -0,0 +1,47 @@
...
+  public static Duration getTimeoutProperty(String propertyName) {
+    var property = System.getProperty(propertyName);
+    if (property == null) {
+      return null;
+    }
+
+    try {
+      return Duration.ofMinutes(Integer.parseInt(property));
+    } catch (NumberFormatException ignored) {
+      return Duration.parse(property);
```
nquinquenel, line 44: There is a `DateTimeParseException` thrown in case it's wrong, shouldn't it be safer to catch it and return `null`?

nquinquenel (approved):
LGTM

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/DurationUtils.java
thahnen: I re-used the old logic that got lost (see [ticket comment](https://sonarsource.atlassian.net/browse/SLE-851?focusedCommentId=604251)) from SLCORE.

I think you're right. I will change it that it is logged and then returns `null`. I wouldn't fail hard in this case but for any means necessary the user should be informed the input was garbage :smile:

PR #667, opened 2024-05-23
SLE-855: Rely on mirrored PyDev

thahnen:
For Cirrus CI rely on the mirrored PyDev Eclipse Update site in order to stabilize the build. The issue is not present locally and if it occurs, Maven can just be called again.


thahnen:
This was handled by mirroring the PyDev Eclipse Update Sites instead of introducing a workaround.

PR #668, opened 2024-05-24
SLE-869: Use Cirrus CI cache for orchestrator

thahnen:
To reduce the network usage with Repox while speeding up the the Cirrus CI pipeline.


nquinquenel (approved):
LGTM

PR #669, opened 2024-05-27
SLE-870: Rely on SLCORE issue tracking

thahnen:
Analyze with issue tracking from SLOCRE, getting rid of the logic on SLE.


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/SonarLintEclipseHeadlessRpcClient.java
```diff
@@ -204,11 +205,27 @@ public void didDetectSecret(String configScopeId) {
   }
 
   @Override
-  public void didRaiseIssue(String configurationScopeId, UUID analysisId, RawIssueDto rawIssue) {
+  public void raiseIssues(String configurationScopeId, Map<URI, List<RaisedIssueDto>> issuesByFileUri, boolean isIntermediatePublication, UUID analysisId) {
     var currentAnalysis = RunningAnalysesTracker.get().getById(analysisId);
-    if (currentAnalysis != null) {
-      currentAnalysis.addRawIssue(rawIssue);
+    if (currentAnalysis == null) {
+      return;
```
damien-urruty-sonarsource, line 211: Please be aware that the override is slightly wrong here, the `analysisId` parameter is `Nullable` in `SonarLintRpcClientDelegate`. This is to accommodate the case where we call `raiseIssues` outside of any analysis, for example when we receive a server-event impacting issues (e.g. an issue was marked as resolved on the server). It's not a problem with SLCORE 10.2 where the `analysisId` should always be non-null, but we are working at the moment on [SLCORE-677](https://sonarsource.atlassian.net/browse/SLCORE-677) which leads to `analysisId`s being null.

I would advise not to rely on `analysisId` here. Raising issues should be independent from the analysis

damien-urruty-sonarsource: We copied `analysisId` from `DidRaiseIssueParams` when we implemented `RaiseIssuesParams` but in hindsight, it should probably not even be present

thahnen: This was now mitigated. We treat those `raiseIssues(...)` calls just as something unexpected from SLCORE ^^

jblievremont (approved):
LGTM 🔥 🚢 🚀

PR #673, opened 2024-06-05
SLE-872: Enhance project preference page

thahnen:
The enhancement includes information when the preference is set for New Code. An undead `serverLink` was removed, that as never used anyway.


jblievremont (approved):
LGTM 👍🏻

PR #682, opened 2024-06-14
fix(BUILD-5327): generic JMESPATH to find cloudfront distribution ID

julien-carsique-sonarsource:
Improve the JMESPATH query to get it more generic (no specific index in the array of origins)


thahnen (approved):
LGTM

PR #686, opened 2024-06-19
SLE-806: Only support SonarQube 9.9 and newer

thahnen:
SonarQube 9.9 is the latest active, long-term version; SonarLint will only support this and all newer versions. The changes will be merged after they are made in SLCORE.

### DIFF
--- a/.cirrus.default.yml
+++ b/.cirrus.default.yml
@@ -172,16 +172,20 @@ qa_task:
   matrix:
     - env:
         TARGET_PLATFORM: 'oldest'
-        SQ_VERSION: 'LATEST_RELEASE[8.9]'
-        QA_CATEGORY: 'Oldest'
+        SQ_VERSION: 'LATEST_RELEASE[9.9]'
+        QA_CATEGORY: 'oldest-LATEST_RELEASE_99'
     - env:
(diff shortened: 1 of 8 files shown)


thahnen:
This error is coming from the [oldest.target](https://github.com/SonarSource/sonarlint-eclipse/blob/10.2.0.81875/its/target-platforms/oldest.target) running on `JavaSE-11` while SQ 9.9 LTS requires Java 17 as a minimum ([latest.target](https://github.com/SonarSource/sonarlint-eclipse/blob/10.2.0.81875/its/target-platforms/latest.target) and [ibuilds.target](https://github.com/SonarSource/sonarlint-eclipse/blob/10.2.0.81875/its/target-platforms/ibuilds.target)) already running on it. Maybe the latter one should be moved to Java 21 at some point.

[Orchestrator_SonarQubeConnectedModeTest-error.log](https://github.com/user-attachments/files/15925492/Orchestrator_SonarQubeConnectedModeTest-error.log)

thahnen:
We have to change the `orchestrator.properties` to use Java 17 as `java.home`

eray-felek-sonarsource (approved):

PR #687, opened 2024-06-20
SLE-888: Split debug logs and IDE-specific traces

thahnen:
## Summary

The **SonarLint Console** of the builtin Console view currently offers the user the option to display the *Verbose output* (containing the debug statements) and the *Analysis logs* next to the default output.

For the last release, I changed the behavior for Eclipse-specific invocations (e.g., adaptations or extension-point interaction) to also log their output that can be used when debugging third-party integrations with SonarLint (e.g., from IBM Developer for z/OS). These logs are very useful but pollute the output for users who want to debug "normal" issues with SonarLint.

See this exemplary screenshot for example, on how much of the output is not coming from either the *Verbose output* (actual debug logs of SonarLint logic) and *Analysis logs*. Colors are OS-specific but differ:

Screenshot from 2024-06-20 14-15-00

In order to keep the normal *Verbose output* simple and SonarLint-logic-specific, a new option, *IDE-specific traces*, was added to the **SonarLint Console** that can be enabled separately when we ask for it. For now, the behavior for asking users for "logs" will be for them to have *Verbose output* and *Analysis logs* enabled, they are distinguished by color.
When we want to go into more detail, as this is not used in most cases, we can ask them to also enable the *IDE-specific traces* option - it is called *tracing* in context of Eclipse IDE.

## Testing

This is already used in the `StandaloneAnalysisTest#test_jdt_java_like_extension_COBOL()` but is not necessary for the other unit/integration tests. Therefore the option is disabled by default.

### DIFF
--- a/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
+++ b/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
@@ -138,9 +138,8 @@ public class StandaloneAnalysisTest extends AbstractSonarLintTest {
   @Test
   public void test_jdt_java_like_extension_COBOL() {
     // INFO: Here we check for the IDE-specific logs, so we have to enable them!
-    var consoleView = new SonarLintConsole();
     try {
-      consoleView.enableIdeSpecificLogs(true);
+      new SonarLintConsole().enableIdeSpecificLogs(true);
 
       new JavaPerspective().open();
       var rootProject = importExistingProjectIntoWorkspace("connected", "connected");
@@ -148,13 +147,20 @@ public class StandaloneAnalysisTest extends AbstractSonarLintTest {
       var cobolFile = rootProject.getResource("Test.cbl");
       openFileAndWaitForAnalysisCompletion(cobolFile);
 
-      // even when no language found, the Secrets analyzer should at least give it a shot^^
-      var consoleText = new SonarLintConsole().getConsoleView().getConsoleText();
-      assertThat(consoleText)
-        .contains("Execute Sensor: TextAndSecretsSensor")
-        .doesNotContain("File 'Test.cbl' excluded by 'JavaProjectConfiguratorExtension'");
+      // To see that the analysis ran and the file wasn't skipped by JDT, we await the "Analysis output" of the
+      // SonarText analyzer starting.
+      // To see that enabling the IDE-specific logs in the SonarLint Console worked, we wait for the adaptation
+      // trace message of the SonarLintPostBuildListener coming in.
+      // To see that the file wasn't excluded by JDT, we also check that that message is not in the logs: It cannot
+      // happen if the SonarText analyzer starting message is coming and vice versa.
+      Awaitility.await()
+        .untilAsserted(() -> assertThat(new SonarLintConsole().getConsoleView().getConsoleText())
+          .contains("Execute Sensor: TextAndSecretsSensor")
+          .contains(
+            "[SonarLintPostBuildListener#visitDelta] Try get project of resource 'L/connected/Test.cbl' -> 'L/connected/Test.cbl' could not be adapted to 'org.sonarlint.eclipse.core.resource.ISonarLintProject'")
+          .doesNotContain("File 'Test.cbl' excluded by 'JavaProjectConfiguratorExtension'"));
     } finally {
-      consoleView.enableIdeSpecificLogs(false);
+      new SonarLintConsole().enableIdeSpecificLogs(false);
     }
   }
 
--- a/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
+++ b/its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
@@ -137,17 +137,25 @@ public class StandaloneAnalysisTest extends AbstractSonarLintTest {
   /** See SLE-854: JDT tries to find files with 'Java-like' extensions even if they're not Java */
   @Test
   public void test_jdt_java_like_extension_COBOL() {
-    new JavaPerspective().open();
-    var rootProject = importExistingProjectIntoWorkspace("connected", "connected");
+    // INFO: Here we check for the IDE-specific logs, so we have to enable them!
+    var consoleView = new SonarLintConsole();
+    try {
+      consoleView.enableIdeSpecificLogs(true);
 
-    var cobolFile = rootProject.getResource("Test.cbl");
-    openFileAndWaitForAnalysisCompletion(cobolFile);
+      new JavaPerspective().open();
+      var rootProject = importExistingProjectIntoWorkspace("connected", "connected");
 
-    // even when no language found, the Secrets analyzer should at least give it a shot^^
-    var consoleText = new SonarLintConsole().getConsoleView().getConsoleText();
-    assertThat(consoleText)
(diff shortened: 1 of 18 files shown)


its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
```diff
@@ -137,17 +137,25 @@ public void analyze_automatic_workspace_build_disabled() {
...
+      var rootProject = importExistingProjectIntoWorkspace("connected", "connected");
+
+      var cobolFile = rootProject.getResource("Test.cbl");
+      openFileAndWaitForAnalysisCompletion(cobolFile);
+
+      // even when no language found, the Secrets analyzer should at least give it a shot^^
+      var consoleText = new SonarLintConsole().getConsoleView().getConsoleText();
+      assertThat(consoleText)
+        .contains("Execute Sensor: TextAndSecretsSensor")
+        .doesNotContain("File 'Test.cbl' excluded by 'JavaProjectConfiguratorExtension'");
```
sophio-japharidze-sonarsource, line 155: no big deal but this test does not really guarantee that the IDE specific logs will work, it checks that the log line will NOT be there - either it's really not there or the logging is not working :D 

If possible, maybe we could also assert that something from the IDE logs IS there

its/src/org/sonarlint/eclipse/its/reddeer/views/SonarLintConsole.java
```diff
@@ -70,6 +70,16 @@ public void enableAnalysisLogs() {
     }
   }
 
+  public void enableIdeSpecificLogs(boolean enabled) {
+    consoleView.activate();
+    var menu = new ToolItemMenuItem(new DefaultToolItem(consoleView.getCTabItem().getFolder(), "Configure logs"), "IDE-specific traces");
+    if (!menu.isSelected() && enabled) {
```
sophio-japharidze-sonarsource, line 76: identical implementation in if-else branches??

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/LogListener.java
```diff
@@ -29,4 +29,10 @@ public interface LogListener {
 
   void debug(@Nullable String msg, boolean fromAnalyzer);
 
+  /**
+   *  This should only be used for IDE-specific logging and is not intended for tracing messages from SLCORE as these
+   *  ones are handled like debug messages. IDE-specific logging is something like adaptations, interaction with an
+   *  extension point, ...
+   */
+  void trace(@Nullable String msg);
```
sophio-japharidze-sonarsource, line 37: I would suggest to rename this method to something more descriptive like `traceIdeMessage`, otherwise it is guaranteed to be confused with SLCORE trace method, even though you have nice comments on top :)

sophio-japharidze-sonarsource (commented):
Please have a look at the comments and LMK in case they don't make sense to you.

its/src/org/sonarlint/eclipse/its/reddeer/views/SonarLintConsole.java
thahnen: No, the `menu.select()` is both used for selecting and de-selecting :face_in_clouds: I can merge the branches tho and add a comment for when it applies:
- we don't want it enabled, but currently it is
- we want it enabled, and currently, it is not

its/src/org/sonarlint/eclipse/its/StandaloneAnalysisTest.java
thahnen: Good idea, I see what a reliable log message we could use

thahnen: Done

its/src/org/sonarlint/eclipse/its/reddeer/views/SonarLintConsole.java
thahnen: Done

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/LogListener.java
thahnen: Done

its/src/org/sonarlint/eclipse/its/reddeer/views/SonarLintConsole.java
```diff
@@ -70,6 +70,18 @@ public void enableAnalysisLogs() {
...
   }
 
+  public void enableIdeSpecificLogs(boolean enabled) {
+    consoleView.activate();
+    var menu = new ToolItemMenuItem(new DefaultToolItem(consoleView.getCTabItem().getFolder(), "Configure logs"), "IDE-specific traces");
+
+    // When the IDE-specific logs are not selected but we want to enable them, we "select" the menu item.
+    // When the IDE-specific logs are selected but we want to disable them, we "de-select" the menu item.
+    // -> `menu.select()` is used for toggling them on/off, not the greatest naming.
+    if ((!menu.isSelected() && enabled) || (menu.isSelected() && !enabled)) {
```
sophio-japharidze-sonarsource, line 80: Is the 'if statement' even needed? maybe the comment above is enough

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/preferences/SonarLintPreferencesInitializer.java
```diff
@@ -32,7 +32,9 @@ public class SonarLintPreferencesInitializer extends AbstractPreferenceInitializ
   public void initializeDefaultPreferences() {
     var node = DefaultScope.INSTANCE.getNode(SonarLintUiPlugin.PLUGIN_ID);
     node.put(SonarLintConsole.P_SHOW_CONSOLE, SonarLintConsole.P_SHOW_CONSOLE_ON_ERROR);
+    node.putBoolean(SonarLintConsole.P_VERBOSE_OUTPUT, false);
```
sophio-japharidze-sonarsource, line 35: What was happening before this change? verbose output was enabled by default?

its/src/org/sonarlint/eclipse/its/reddeer/views/SonarLintConsole.java
thahnen: Yes it is, so we don't accidentily trigger the menu item on or off. As I said it can be both selecting and de-selecting.
This if statement is used so we only select it if we really want to have it enabled (if already enabled, nothing needs to be done) and we de-select it if we want to have it disabled (if already disabled, nothing needs to be done).

sophio-japharidze-sonarsource (approved):

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/preferences/SonarLintPreferencesInitializer.java
thahnen: This is not linked to the UI. This is for the preferences saved on disk as Eclipse does preserve them this way.
E.g. when exporting the preferences of the workspace (the settings) to share them with others, this is used to fill in the default values.
When the other one was added some time ago, I assume this was just forgotten.

PR #689, opened 2024-06-21
SLE-879: Java analyzer update + new plug-in

thahnen:
The SonarJava analyzer now consists of two plug-ins: The already existing `sonar-java-plugin` and the new `sonar-java-symbolic-execution-plugin`.


sophio-japharidze-sonarsource (approved):
have not tested if it works

thahnen:
@sophio-japharidze-sonarsource tested in the ITs against SQ 8.9, 9.9, 10.6 DEV and tested manually against 9.9 LTS while trying to break things :smile: Nothing broke :disappointed:

PR #694, opened 2024-06-28
SLE-892: Support of Eclipse Buildship (Gradle)

thahnen:
## Summary

Some time ago we [introduced](https://github.com/SonarSource/sonarlint-eclipse/commit/d8f408823e4348b25e8ba38f495f8c416f5fac50) the Shared Connected Mode configuration files to be shared among contributors. For this, an additional extension point had to be introduced for the concept of hierarchy that was not available in Eclipse and is only emulated in a very small scope by plug-ins like [m2eclipse](https://projects.eclipse.org/projects/technology.m2e).

This finally offered the possibility to introduce a better integration with hierarchy-based project structures introduced by build systems like Maven and Gradle.

This means that support for Gradle via the official [Eclipse Buildship](https://projects.eclipse.org/projects/tools.buildship) plug-in is finally coming to SonarLint for Eclipse. Currently, the feature is only used for exporting and consuming Shared Connected Mode files but it leads to less friction for setting up Connected Mode for one-to-many projects in the workspace!

## Testing

Testing that the Gradle integration works is done via non-Connected Mode integration tests. The actual connection/binding is independent of it and won't be featured here. Mostly manually tested for now (and for when the Shared Connected Mode was implemented in the first place, some time ago). See the ticket content for more information.


jblievremont (approved):
Nice! Tested manually in Eclipse 2024-06 with the latest buildship plugin; and nice work on the integration test!

PR #695, opened 2024-07-02
NO-JIRA: Don't run unit tests again for SonarQube

thahnen:
This is primarily used to not be blocked if unit tests decide to be flaky here instead of the actual unit-test-invoking "validate" task. On the other hand it speeds up the build by a minute or so.

### DIFF
--- a/.cirrus.default.yml
+++ b/.cirrus.default.yml
@@ -267,7 +267,7 @@ sonarqube_task:
     unzip jacoco.zip -d ${CIRRUS_WORKING_DIR}/org.sonarlint.eclipse.core.tests/target/
     source cirrus-env QA
     .cirrus/regular_mvn_build_deploy_analyze \
-      -P-deploy-sonarsource,-release,-sign -Dmaven.install.skip=true -Dmaven.deploy.skip=true \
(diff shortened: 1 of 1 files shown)


.cirrus.default.yml
```diff
@@ -270,7 +270,7 @@ sonarqube_task:
     unzip jacoco.zip -d ${CIRRUS_WORKING_DIR}/org.sonarlint.eclipse.core.tests/target/
     source cirrus-env QA
     .cirrus/regular_mvn_build_deploy_analyze \
-      -P-deploy-sonarsource,-release,-sign -Dmaven.install.skip=true -Dmaven.deploy.skip=true \
+      -P-deploy-sonarsource,-release,-sign -Dmaven.install.skip=true -DskipTests -Dmaven.deploy.skip=true \
```
jblievremont, line 273: For consistency, may I suggest using `maven.test.skip=true` 😇?

thahnen: We cannot use it, as it is for Tycho Surefire, not Maven :open_mouth:

jblievremont (approved):
LGTM!

PR #701, opened 2024-07-08
MMF-3807: Raise awareness about recent changes and newly released versions

thahnen:
## Summary

Currently, Eclipse users don't know what has changed when updating SonarLint except for taking a look at the GitHub release page or the Sonar Community forum announcement. Additionally, they're not aware if a new version of SonarLint for Eclipse was released except if they trigger the **Check for Updates** action in Eclipse.
And even if users did not install it from our own Update Site (or the marketplace that links to it), they won't be informed about new versions.

We want to change that by raising awareness about what has changed by providing the Release Notes directly inside the IDE and informing the user after a fresh installation/update. For newer versions available, we check once per day against our own Update Site, and in case a newer version is available, we inform the user as well, encouraging them to update.

## Testing

This behavior only happens once the IDE loads, and if a newer version is available for the new version check, we cannot test that via integration tests for now. We can only check the new Release Notes preference page. Even the Welcome view we edited cannot be tested as this is a web browser and its "paging" cannot be interacted with using RedDeer.


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/http/HttpClient.java
```diff
@@ -0,0 +1,187 @@
...
+import javax.net.ssl.TrustManagerFactory;
+import org.eclipse.jdt.annotation.Nullable;
+import org.sonarlint.eclipse.core.SonarLintLogger;
+import org.sonarsource.sonarlint.core.rpc.protocol.backend.initialize.HttpConfigurationDto;
+
+/**
+ *  Very bare bones implementation of a HTTP client that can work in the context of SonarLint both with a configurable
+ *  key and trust store. Additionally, timeouts can be provided that will used the same way as SLCORE would use them.
+ */
+public class HttpClient {
```
nquinquenel, line 45: It seems like reinventing the wheel a lot. Couldn't you leverage the existing HTTP client in SLCORE?

thahnen: Sadly no :cry:

Before 10.0, it was still available in the `org.sonarsource.sonarlint.core.clientapi.SonarLintBackend` but only to open HTTP connections to a specific connection. Since 10.0 it was completely moved to the part out of process, not being available to the clients anymore.

I thought about sparking a conversation with the Sonar Solution squad about it, but in the end decided against it for the following reasons:
- having an HTTP client as API would enable (malicious) third parties to plug into that and use it to trigger calls on behalf of SonarLint (Core), e.g. when SonarLint is the only plug-in configured with HTTP/SSL configurations and others are not
- in general, no client should make any HTTP requests as everything they need is either in SLCORE or provided by the IDE, here this is a solution as Eclipse doesn't provide what is needed (and probably never will be) - information about the Release Notes as well as newly released versions
- the implementation is as bare-bones as it gets, I was even thinking in the first place of dropping the `SSLContext` related logic but keeping it otherwise might have rendered this feature useless - the HTTP/SSL configuration provided by users is there for a reason in the first place

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/SonarLintUiPlugin.java
```diff
@@ -259,8 +265,38 @@ public IStatus run(IProgressMonitor monitor) {
...
+        if (!SonarLintGlobalConfiguration.sonarLintVersionHintHidden()
+          && !SonarLintGlobalConfiguration.isSonarLintVersionHintDateToday()) {
+          var newestSonarLintVersion = EclipseUpdateSite.getNewestVersion();
+          var currentSonarLintVersion = BundleUtils.getBundleVersion();
+          SonarLintLogger.get().debug("Current version (" + currentSonarLintVersion + "), newest version ("
+            + newestSonarLintVersion + ")");
+
+          if (newestSonarLintVersion == null) {
+            SonarLintLogger.get().debug("Cannot check for newer SonarLint versions via the SonarLint for Eclipse "
+              + "Update Site. In thase case, please check the Sonar Community Forum ("
```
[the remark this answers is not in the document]
thahnen: Good catch, will fix that later

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/http/HttpClient.java
nquinquenel: I was not thinking about using the HTTP client in SLE directly, but instead creating a dedicated service in SLCORE. This could be, perhaps, reused in other flavors.

nquinquenel (approved):
Not a huge fan of the HTTP service but let's go like that for now

PR #702, opened 2024-07-11
SLE-898: Enable DBD in Connected Mode

thahnen:
Dataflow bug detection for Java and Python is available since SQ 10.6 and in SC for SonarLint to be used.


nquinquenel (approved):
LGTM

PR #704, opened 2024-07-15
SLE-764: Simplified SonarQube analysis

thahnen:
SonarQube analysis will not contain **-SNAPSHOT** anymore and only the **{Major}.{Minor}.{Patch}** version; this will simplify the release process as well, not having to ask a champion to change the version on Next.

On SLE, we only have master, maintenance branches, and pull requests, removing unused code. The version on all available branches, including master and maintenance branches, looks like **{Major}.{Minor}.{Patch}-SNAPSHOT**


henryju (approved):
LGTM

PR #705, opened 2024-07-15
SLE-901: Change project name check

thahnen:
As Eclipse PDE decided to change the project name.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/DefaultSonarLintProjectsProvider.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/DefaultSonarLintProjectsProvider.java
@@ -52,7 +52,7 @@ public class DefaultSonarLintProjectsProvider implements ISonarLintProjectsProvi
   /** This should be extended with further checks from other plug-ins providing technical projects */
   private static boolean isTechnicalProject(String projectName) {
     // check for Eclipse PDE technical projects
-    var isTechnicalProject = TECHNICAL_PROJECT_PDE.equals(projectName) && Platform.getBundle("org.eclipse.pde.ui") != null;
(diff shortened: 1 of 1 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/DefaultSonarLintProjectsProvider.java
```diff
@@ -52,17 +52,17 @@ public Collection<ISonarLintProject> get() {
   /** This should be extended with further checks from other plug-ins providing technical projects */
   private static boolean isTechnicalProject(String projectName) {
     // check for Eclipse PDE technical projects
-    var isTechnicalProject = TECHNICAL_PROJECT_PDE.equals(projectName) && Platform.getBundle("org.eclipse.pde.ui") != null;
+    var isTechnicalProject = projectName.contains(TECHNICAL_PROJECT_PDE) && Platform.getBundle("org.eclipse.pde.ui") != null;
 
     // check for Eclipse PDT technical projects
     isTechnicalProject = isTechnicalProject ||
       (TECHNICAL_PROJECT_PDT.equals(projectName) && Platform.getBundle("org.eclipse.php.ui") != null);
```
nquinquenel, line 59: What about PDT?

nquinquenel (approved):
1 question

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/resources/DefaultSonarLintProjectsProvider.java
thahnen: The technical project is only called `Servers` and if we just check if a project name contains this, this will rule out more projects by accident I think. The other ones are called so cryptically, that I doubt anyone will call their project like this, also because it might not even work.

PR #711, opened 2024-07-30
SLE-798: Add IT for custom secrets

thahnen:
## Summary

In order to test custom secrets, the Sonar WS artifact had to be updated, with slight adjustments to some tests. Additionally, the move to SonarQube Enterprise from the orchestrator had to be done as the custom secrets are only available there.


nquinquenel (approved):
LGTM!

PR #715, opened 2024-08-06
SLE-873: Split ITs into separate bundles

thahnen:
## Summary

With these changes, the integration tests are split up to speed up the build and distinguish between the use cases. This also enables us to lower the flakiness of some tests that are based on the Eclipse IDE version used. The tests against SonarCloud are not yet active and will be done with SLE-911.

The target platforms were overhauled towards more different tests based on the Eclipse IDE version (some bundle Java 11, some 17 or 21). We still build against the latest Eclipse IDE with Java 11 (`latest-java-11_e424.target`). The IDE configuration now correctly uses the latest Java 17 dependencies so less configuration by the developer (which is me).

In order to make the tests parallelizable, some black magic had to be done to `its/pom.xml` with profiles that can be deactivated to limit the scope of the ITs (e.g., you want only non-Connected Mode tests, then disable Connected Mode tests with SQ / SC, ...). This had to be done to still work correctly with Tycho and the Eclipse IDE itself.

## Testing

The Cirrus CI pipeline was changed and basically is the *testing* of this effort as no changes on production code or the ITs (except for fixing paths that have changed or names of the target platforms) have been done.


.cirrus.default.yml
```diff
@@ -149,7 +149,7 @@ validate_task:
   eks_container:
     <<: *BUILDER_CONTAINER_DEFINITION
     cpu: 4
-    memory: 8G
```
nquinquenel, line 152: Just to double check - reducing the spec is on purpose?

.cirrus.default.yml
```diff
@@ -242,8 +240,151 @@ qa_task:
...
+      path: "**/reports/**/*"
+    junit_artifacts:
+      paths:
+      - "**/test-results/**/*.xml"
+      - "**/target/surefire-reports/TEST-*.xml"
+      format: junit
+    slack_notification_script: |
+      source slack-failure-notification "Cirrus CI build failure on pipeline: $CIRRUS_PIPELINE_NAME"
+
+#qa_connectedModeSonarCloud_task:
```
nquinquenel, line 262: I wouldn't keep this as commented, even if it's not yet enabled. Perhaps you could use `skip` or `only_if` and directly set the boolean

.cirrus.default.yml
```diff
@@ -242,8 +240,151 @@ qa_task:
...
+
+qa_standaloneMode_task:
+  #  name: "Run ITs for Standalone Mode"
+  depends_on:
+    - build
+  <<: *ONLY_IF_EXCEPT_NIGHTLY
+  eks_container:
+    <<: *BUILDER_CONTAINER_DEFINITION
+    cpu: 4
+    memory: 10G
```
nquinquenel, line 336: Nitpick: standalone have 10G vs 12G for connected mode, is it expected?

.cirrus.default.yml
```diff
@@ -229,11 +222,16 @@ qa_task:
     echo 'Recording tests on video'
     ffmpeg -loglevel warning -f x11grab -video_size 1920x1080 -i ${DISPLAY} -codec:v libx264 -r 12 ${CIRRUS_WORKING_DIR}/recording_${QA_CATEGORY}.mp4
   run_its_script: |
-    echo "Run Maven ITs for Eclipse ${TARGET_PLATFORM} and Server ${SQ_VERSION}"
-    mvn -B -e -V org.jacoco:jacoco-maven-plugin:prepare-agent verify -f its/pom.xml -Pcoverage \
-      -Dtarget.platform=${TARGET_PLATFORM} -Dtycho.localArtifacts=ignore -Dsonarlint-eclipse.p2.url="file://${CIRRUS_WORKING_DIR}/staged-repository" -Dsonar.runtimeVersion=${SQ_VERSION} \
+    echo "Run Maven ITs for Connected Mode with SonarQube on Eclipse 2024-03 / 4.31 (latest Java 17) and Server ${SQ_VERSION}"
+    cd its/
+    mvn -B -e -V org.jacoco:jacoco-maven-plugin:prepare-agent verify \
+      -P coverage,\!standaloneMode,\!connectedModeSc \
```
nquinquenel, line 228: `\!standaloneMode,\!connectedModeSc` what does this mean?

.cirrus.default.yml
```diff
@@ -242,8 +240,151 @@ qa_task:
...
+    set -euo pipefail
+    /etc/init.d/xvfb start
+    metacity --sm-disable --replace &
+    sleep 10 # give metacity some time to start
+    echo 'Recording tests on video'
+    ffmpeg -loglevel warning -f x11grab -video_size 1920x1080 -i ${DISPLAY} -codec:v libx264 -r 12 ${CIRRUS_WORKING_DIR}/recording_${TARGET_PLATFORM}.mp4
+  run_its_script: |
+    echo "Run Maven ITs for Standalone Mode on Eclipse '${TARGET_PLATFORM}'"
+    cd its/
+    mvn -B -e -V org.jacoco:jacoco-maven-plugin:prepare-agent verify \
```
nquinquenel, line 367: Do you really need to run jacoco during ITs?

its/README.md
```diff
@@ -0,0 +1,81 @@
...
+
+```
+cd $SONARLINT_FOR_ECLIPSE_FOLDER
+mvn clean verify -DskipTests
+cd its/
+mvn clean verify -Dtarget.platform=latest-java-17_e431 -Dtycho.localArtifacts=ignore -Dsonarlint-eclipse.p2.url="file://{path to repo}/org.sonarlint.eclipse.site/target/repository" -DskipTests
+```
+
+When all the projects are opened inside Eclipse and the target platform correctly configured, this is done in the
+background (not for the main SonarLint for Eclipse project tho if not also imported into Eclipse, which is required).
```
nquinquenel, line 28:
```suggestion
background (not for the main SonarLint for Eclipse project though if not also imported into Eclipse, which is required).
```

its/org.sonarlint.eclipse.its.connected.sc/build.properties
```diff
@@ -0,0 +1,4 @@
+source.. = src/
+output.. = target/classes/
+bin.includes = META-INF/,\
+               .
```
nquinquenel, line 4: Strange formatting

.cirrus.default.yml
```diff
@@ -242,8 +240,151 @@ qa_task:
...
+    type: m6a.large
+  env:
+    ARTIFACTORY_API_KEY: VAULT[development/artifactory/token/${CIRRUS_REPO_OWNER}-${CIRRUS_REPO_NAME}-private-reader access_token]
+    GITHUB_TOKEN: VAULT[development/github/token/licenses-ro token]
+    SONARCLOUD_IT_PASSWORD: VAULT[development/team/sonarlint/kv/data/sonarcloud-it data.password]
+    DISPLAY: :10
+    MAVEN_OPTS: -Xmx3072m
+  matrix:
+    - env:
+        TARGET_PLATFORM: 'oldest-java-11_e48'
```
nquinquenel, line 346: Can't it be `oldest-java-11` instead of `oldest-java-11-e48`? Especially that you use this value later in the test classes, it feels really specific

nquinquenel (approved):
Looks mostly good to me, few comments

.cirrus.default.yml
thahnen: Yes, I've checked the statistics and DevInfra asked to lower the specs if they're not needed. At most I used 2GB and sometimes it spikes to 3GB and therefore 4GB is enough.

.cirrus.default.yml
thahnen: I will immediately start working on this, that's why it is still in here :smile:

.cirrus.default.yml
thahnen: Yes, because no Orchestrator is running on this and does not require that much RAM

.cirrus.default.yml
thahnen: Disable the profiles, this is the black magic I was talking about. In order to work correctly with Maven + Tycho + inside the Eclipse IDE we have to have everything available and enabled, but when we run on the CI/CD we cannot run all the tests together and will therefore also skip the modules completely (compilation and so on).

.cirrus.default.yml
thahnen: Yes, on Eclipse 99% of the coverage depend on the ITs as 99% of the tests cannot run on unit tests alone because they need an Eclipse instance. So don't be confused when comparing it to IntelliJ, we have no light or heavy or even medium tests as it is not possible :sob:

.cirrus.default.yml
thahnen: Nope, the `e48` references to Eclipse 4.8 which is Photon. I've decided to add the name (it will forever be the oldest Java 11 supporting Eclipse IDE) in order to easier find issues.

its/org.sonarlint.eclipse.its.connected.sc/build.properties
thahnen: Yes for Eclipse this is normal :smile: and required :face_in_clouds:

.cirrus.default.yml
thahnen: The other ones reference to the specific Eclipse versions as well: `e424` is the latest Eclipse IDE shipping with Java 11, `e431` the latest with Java 17 and for Java 21 this is currently done therefore no added Eclipse version.

PR #717, opened 2024-08-12
SLE-911: Implement first test with SonarCloud

thahnen:
Reworked the old (and never used) IT and embedded it into the new environment.


nquinquenel (approved):
LGTM

PR #719, opened 2024-08-14
SLE-857: Fix flaky ITs with SQ

thahnen:
Due to timing issues between the connection and binding wizard some tests were sporadically failing. This is the same for polluted logs.


eray-felek-sonarsource (approved):
LGTM

PR #721, opened 2024-08-15
SLE-846: IT for Maven/Gradle and multiple configurations

thahnen:
## Summary

Since we now also provide support for Gradle, add a test for the Shared Connected Mode configuration based on Maven/Gradle as a hierarchical project provider and two separate JSON files.

Also, add a test for exporting the shared Connected Mode configuration from SonarQube.


eray-felek-sonarsource (approved):
LGTM

PR #723, opened 2024-08-19
SLE-920: Enhance the extension point for diff viewers

thahnen:
Enhance the extension point used for syntax highlighting with a new method to gather language (aka plug-in) -specific implementations of the `TextMergeViewer` that is used for displaying diffs.

Added implementations for JDT (Java/JSP) and CDT (C/C++) but not PyDev due to a missing `Export-Package` statement in the plug-in.

For the latter one, there is a related PR opened in https://github.com/fabioz/Pydev/pull/370 but this won't be used in the feature implementation for now, but can be used later!


org.sonarlint.eclipse.cdt/META-INF/MANIFEST.MF
```diff
@@ -14,7 +14,9 @@ Require-Bundle: org.eclipse.core.runtime,
  org.sonarlint.eclipse.core,
  org.eclipse.core.filesystem,
  org.eclipse.jdt.annotation;resolution:=optional,
- org.eclipse.text
+ org.eclipse.text,
```
sophio-japharidze-sonarsource, line 17: what do these lines do? just curious

org.sonarlint.eclipse.cdt/src/org/sonarlint/eclipse/cdt/internal/CProjectConfiguratorExtension.java
```diff
@@ -103,17 +115,26 @@ public SonarLintLanguage language(ISonarLintFile file) {
...
+    if (isCdtUiPresent() && (ruleLanguage.equals(C_LANGUAGE_KEY) || ruleLanguage.equals(CPP_LANGUAGE_KEY))) {
       return Optional.of(CdtUiUtils.sourceViewerConfiguration());
     }
     return Optional.empty();
   }
 
   @Override
   public Optional<IDocumentPartitioner> documentPartitioner(String ruleLanguage) {
-    if (isCdtPresent() && (ruleLanguage.equals(C_LANGUAGE_KEY) || ruleLanguage.equals(CPP_LANGUAGE_KEY))) {
+    if (isCdtUiPresent() && (ruleLanguage.equals(C_LANGUAGE_KEY) || ruleLanguage.equals(CPP_LANGUAGE_KEY))) {
```
sophio-japharidze-sonarsource, line 126: I feel like you could export this check into a separate function

org.sonarlint.eclipse.cdt/META-INF/MANIFEST.MF
thahnen: It is basically the Eclipse way of defining dependencies at runtime (and therefore also compile time).

The `MANIFEST.MF` and specifically this section will contain all bundles that we require at the runtime of the Eclipse installation. Most of them are present due to them being part of the Eclipse platform itself, but for some we have an optional.

E.g. here we have the `org.eclipse.cdt.core;resolution:=optional` statement that more or less is used for us to only make use of this very sub-plugin if CDT (the C development tools) are installed.

What I added is the compare bundle that brings in the `TextMergeViewer` and SWT that is the UI system of Eclipse.

org.sonarlint.eclipse.cdt/src/org/sonarlint/eclipse/cdt/internal/CProjectConfiguratorExtension.java
thahnen: Done

sophio-japharidze-sonarsource (approved):

PR #724, opened 2024-08-20
SLE-921: Dialog displayed for each suggestion change

thahnen:
## Summary

To open a fix suggestion from SonarCloud a lot of logic from the "Open in IDE" feature (independent of SonarQube or SonarCloud) was re-used. Additionally, contrary to the other feature, we don't have to check for branches (this is done already on the SLCORE side) and await the analysis getting "ready" after binding.

When the suggestion is found, the user is offered to accept or decline it or to cancel the whole action, and when the suggestion cannot be found, we inform them in case they still want to know what was suggested and maybe re-use this suggestion by copying it.

## Testing

Integration tests will be created by SLE-923 in order to parallelize it with validation and fix the annoying SLE-922 before that!


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/SonarLintBackendService.java
```diff
@@ -190,7 +190,7 @@ protected IStatus run(IProgressMonitor monitor) {
               Map.of()),
             httpConfiguration,
             getSonarCloudAlternativeEnvironment(),
-            new FeatureFlagsDto(true, true, true, true, false, true, true, true, telemetryEnabled, false),
+            new FeatureFlagsDto(true, true, true, true, false, true, true, true, telemetryEnabled, true),
```
sophio-japharidze-sonarsource, line 193: 🤩

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/StringUtils.java
```diff
@@ -189,4 +189,31 @@ public static String urlDecode(String toDecode) {
...
+    if (!str.contains(lineSeperator)) {
+      return 0;
+    }
+
+    var preSubString = str.substring(0, subStringIndex);
+    return getNumberOfLines(preSubString);
+  }
+
+  public static int getNumberOfLines(final String str) {
+    return (str.length() - str.replace(lineSeperator, "").length()) / lineSeperator.length();
```
sophio-japharidze-sonarsource, line 217: Possible, easier-to-understand alternative

```
str.split(lineSeparator).length()
```
would it miss some cases?

thahnen: I wasn't even thinking about something as simple as this :face_in_clouds: It also shouldn't be missing anything as in case of no line separator in the string, there will be just one element in the list so length will be 1.

I also just noticed the other way around that when I refactored the upper method and moved this to a separate one I forgot to also move the null check in order to not divide by zero.
Once your review is done, I will take this into account and change it :smile:

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/dialog/FixSuggestionUnavailableDialog.java
```diff
@@ -0,0 +1,60 @@
...
+  @Override
+  protected void createButtonsForButtonBar(Composite parent) {
+    createButton(parent, IDialogConstants.OK_ID, "Proceed", true);
+    createButton(parent, IDialogConstants.CANCEL_ID, "Cancel", false);
+  }
+
+  @Override
+  protected void addLabel(Composite container) {
+    var label = new CLabel(container, SWT.WRAP);
+    label.setText("This change suggestion is not applicable as the current code cannot be found in the file.");
```
sophio-japharidze-sonarsource, line 55: "current code on the server"

sophio-japharidze-sonarsource (approved):

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/dialog/AbstractFixSuggestionDialog.java
```diff
@@ -0,0 +1,187 @@
...
+   */
+  protected AbstractFixSuggestionDialog(Shell parentShell, @Nullable SonarLintLanguage language, String explanation,
+    String beforeText, String afterText, int changeIndex, int absoluteNumberOfChanges) {
+    super(parentShell);
+
+    this.language = language;
+    this.explanation = explanation;
+    this.beforeText = beforeText;
+    this.afterText = afterText;
+    this.changeIndex = changeIndex;
```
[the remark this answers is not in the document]
thahnen: Aye aye Captain :wave:

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/dialog/FixSuggestionUnavailableDialog.java
thahnen: Ah I think I still had it before we decided to also update the text merge viewer header. I will change this!

PR #741, opened 2024-09-19
SLE-855: Rely on mirrored PyDev

thahnen:
[SLE-855](https://sonarsource.atlassian.net/browse/SLE-855)

For Cirrus CI rely on the mirrored PyDev Eclipse Update site in order to stabilize the build. The issue is not present locally and if it occurs, Maven can just be called again.


thahnen:
@jayadeep-km-sonarsource the "/" is not a typo. Repository locations in Eclipse PDE target platforms should have a trailing slash, but I'm not consistent with that on all target platforms myself. It is more of a failsafe mechanism.

thahnen:
To be removed with SLE-948

PR #743, opened 2024-09-20
SLE-943: Java analyzer update

thahnen:
(╯°□°)╯︵ ┻━┻


thahnen:
Don't wonder, the commit message is incorrect. I will squash anyway and then it is using the PR title that is correct.

sophio-japharidze-sonarsource (approved):

PR #768, opened 2024-11-20
SLE-910: Remove username/password authentication

thahnen:
To still work on existing username/password based connections, the background logic still exists. This only removes the wizard pages for selecting the authentication type and the username/password authentication page that was already deprecated a year ago.

But when an old username/password-based connection is used (from before this update) and edited, only the option to input a token is available (and therefore, this method of authentication can be switched to). We correctly handle the switch to a token-based authentication model.


nquinquenel (approved):
LGTM

Can you make sure with our PM that we now can effectively remove username/password auth?

PR #791, opened 2025-01-03
SLE-972: Rely on Eclipse WWD embedded Node.js if possible

thahnen:
When the user has no Node.js installed but is using the Eclipse Wild Web Developer tools that come bundled with Node.js, we rely on that specific (and compatible) version instead of forcing the user to set/install a Node.js version.


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/nodejs/NodeJsService.java
```diff
@@ -0,0 +1,64 @@
...
+
+public class NodeJsService {
+  /**
+   *  This has to be on par with the
+   *  - minimal requirement of the embedded SonarJS analyzer (META-INF/MANIFEST.MF -> "NodeJs-Min-Version")
+   *  - bundled Node.js version in Eclipse WWD at https://github.com/eclipse-wildwebdeveloper/wildwebdeveloper
+   *    -> org.eclipse.wildwebdeveloper.node/META-INF/MANIFEST.MF -> "Bundle-SymbolicName"
+   *    -> org.eclipse.wildwebdeveloper.node.win32.x86_64/nodejs-info.properties -> "archiveURL"
+   */
+  private static final boolean IS_NODEJS_EMBEDDER_BUNDLE_AVAILABLE = BundleUtils.isBundleInstalledWithMinVersion(
```
sophio-japharidze-sonarsource, line 35: typo?

sophio-japharidze-sonarsource (approved):
LGTM!
We need to make sure this is properly documented either [here](https://docs.sonarsource.com/sonarqube-for-ide/eclipse/getting-started/requirements/#overview) or somewhere else.

PR #799, opened 2025-01-21
SLE-1047: Delete deprecated EP and config entry

thahnen:
The EP was deprecated long ago and is not used, not even by IBM in their own integration.

The configuration for `moduleKey` can now also be removed as updating from SLE 3.7 to 11 is not possible anyway. Remove its unit test as well.


nquinquenel (approved):
LGTM!

PR #805, opened 2025-01-29
SLE-1050: Remove binding properties listener appropriately

thahnen:
[SLE-1050](https://sonarsource.atlassian.net/browse/SLE-1050)

This was causing errors from SLCORE as the preferences were receiving events after the project was removed.

[SLE-1050]: https://sonarsource.atlassian.net/browse/SLE-1050?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/ConfigScopeSynchronizer.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/ConfigScopeSynchronizer.java
@@ -68,6 +68,7 @@ public class ConfigScopeSynchronizer implements IResourceChangeListener {
           + "' (pre close)");
       if (project != null) {
         SonarLintLogger.get().debug("Project about to be closed: " + project.getName());
+        SonarLintProjectConfigurationManager.removePreferenceChangeListenerForBindingProperties(project);
         backend.getConfigurationService()
           .didRemoveConfigurationScope(new DidRemoveConfigurationScopeParams(getConfigScopeId(project)));
       }
@@ -77,6 +78,7 @@ public class ConfigScopeSynchronizer implements IResourceChangeListener {
           + "' (pre delete)");
       if (project != null) {
         SonarLintLogger.get().debug("Project about to be deleted: " + project.getName());
+        SonarLintProjectConfigurationManager.removePreferenceChangeListenerForBindingProperties(project);
         backend.getConfigurationService()
           .didRemoveConfigurationScope(new DidRemoveConfigurationScopeParams(getConfigScopeId(project)));
       }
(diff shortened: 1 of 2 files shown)


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/ConfigScopeSynchronizer.java
```diff
@@ -77,6 +77,7 @@ public void resourceChanged(IResourceChangeEvent event) {
           + "' (pre delete)");
       if (project != null) {
         SonarLintLogger.get().debug("Project about to be deleted: " + project.getName());
+        SonarLintProjectConfigurationManager.removePreferenceChangeListenerForBindingProperties(project);
```
damien-urruty-sonarsource, line 80: Shouldn't you do the same PRE_CLOSE?

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/preferences/SonarLintProjectConfigurationManager.java
```diff
@@ -54,15 +57,22 @@ public class SonarLintProjectConfigurationManager {
...
-            listener.accept(project);
-          }
-        });
-      });
+      .ifPresent(node -> node.addPreferenceChangeListener(projectBindingPropertiesListener.get(project)));
+  }
+
+  public static void removePreferenceChangeListenerForBindingProperties(ISonarLintProject project) {
+    ofNullable(project.getScopeContext().getNode(SonarLintCorePlugin.PLUGIN_ID))
+      .ifPresent(node -> node.removePreferenceChangeListener(projectBindingPropertiesListener.remove(project)));
```
damien-urruty-sonarsource, line 75: What if `remove` returns null? Should we check that? Or `removePreferenceChangeListener` can accept Nullable values? Maybe this cannot happen, but just to be on the safe-side we could check

damien-urruty-sonarsource (commented):
I have a doubt about the removal of the listener, should it be called from other places too?

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/backend/ConfigScopeSynchronizer.java
thahnen: Good catch! Due to that I will address the other comment as well!

damien-urruty-sonarsource (approved):
LGTM!

PR #815, opened 2025-02-14
SLE-1064: Notification on invalid credentials

thahnen:
[SLE-1064](https://sonarsource.atlassian.net/browse/SLE-1064)

A new notification for users having an invalid token. This way less error messages are shown to the user and SonarLint CORE won't try to reach out for connections that do not work anyway until the user changes the credentials (token).

[SLE-1064]: https://sonarsource.atlassian.net/browse/SLE-1064?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/popup/InvalidTokenPopup.java
+++ b/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/popup/InvalidTokenPopup.java
@@ -1,0 +1,87 @@
+/*
+ * SonarLint for Eclipse
+ * Copyright (C) 2015-2025 SonarSource SA
+ * sonarlint@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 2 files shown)


org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/popup/InvalidTokenPopup.java
```diff
@@ -0,0 +1,87 @@
...
+  @Override
+  protected String getPopupShellTitle() {
+    var message = facade.isSonarCloud() ? "SonarQube Cloud" : "SonarQube Server";
+    return message + " - Invalid token for connection";
+  }
+
+  @Override
+  protected Image getPopupShellImage(int maximumHeight) {
+    return facade.isSonarCloud()
+      ? SonarLintImages.SONARCLOUD_SERVER_ICON_IMG
```
nquinquenel, line 68: I think the name is wrong, it should be SONARQUBE_CLOUD_ICON_IMG 😄

thahnen: Yes, but this was not necessary to to be refactored back then. The next time I touch it all, I will probably refactor all of the names. But for now we keep the OG names
- *this was written by a representative of the OG SonarLint gang* :laughing:

PR #824, opened 2025-02-20
SLE-1078 allow manual connection and import of US SQC config

sophio-japharidze-sonarsource:
[SLE-1078](https://sonarsource.atlassian.net/browse/SLE-1078)

[SLE-1078]: https://sonarsource.atlassian.net/browse/SLE-1078?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/SonarLintUtils.java
```diff
@@ -81,6 +81,14 @@ public static String getSonarCloudUrl() {
     return System.getProperty("sonarlint.internal.sonarcloud.url", "https://sonarcloud.io");
   }
 
+  public static String getSonarCloudUrl(@Nullable String sonarCloudRegion) {
+    if (sonarCloudRegion != null && sonarCloudRegion.equalsIgnoreCase("US")) {
+      return "https://us.sonarcloud.io";
```
thahnen, line 86: Maybe for this we can later introduce the same behavior as for the normal SonarQube Cloud URL: Have a system property `sonarlint.internal.us.sonarcloud.url`. The property should be on par with the one used for SLCORE in `SonarLintBackendService`.

This way we can use them in the launcher/Maven configuration for the ITs.

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/ProjectSuggestionDto.java
```diff
@@ -38,4 +38,5 @@ public ISonarLintProject getProject() {
   public boolean getIsFromSharedConfiguration() {
     return this.isFromSharedConfiguration;
   }
+
```
thahnen, line 41: Maybe get rid of this change to not have it in this files Git history.

thahnen (approved):
LGTM from my side, just a few comments. I excluded the logic from the commit about the Dogfooding environment variable as this will be removed/reverted anyway.

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/utils/SonarLintUtils.java
sophio-japharidze-sonarsource: Yes, exactly. The work is planned for the next iteration in SLCORE

thahnen: Just to put it here: When this is done on SLCORE side, it has to be added here as well as the launchers!

thahnen (approved):
LGTM!

PR #848, opened 2025-03-20
SLE-1105: Use Eclipse RedDeer fork

thahnen:
Use the SonarSource fork of Eclipse RedDeer in our build and remove the workaround solution that has been put in place.

For the oldest ITs we rely on the 4.2.1 version of our fork, for all other Target Platforms we rely on 4.7.1


nquinquenel (approved):
Well done!

PR #849, opened 2025-03-20
SLE-1122: Don't pollute Sentry when running UTs/ITs

thahnen:
[SLE-1122](https://sonarsource.atlassian.net/browse/SLE-1122)

Sentry should only gather data on production environments and should not be polluted by UTs/ITs (running locally or on the CI).

[SLE-1122]: https://sonarsource.atlassian.net/browse/SLE-1122?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nquinquenel (approved):
I think you could have a flag similar to the telemetry one to enable or disable the monitoring based on this during the initialization of SLOOP, rather than modifying the dogfooding env variable?

PR #856, opened 2025-03-25
SLE-1138: EP to load Connected Mode tokens

thahnen:
[SLE-1138](https://sonarsource.atlassian.net/browse/SLE-1138)

### Summary

We currently rely only on Eclipse Equinox Security which is not available on some Eclipse-based IDEs and there will be a move to Java 17 while we still rely on Java 11. These users cannot work with Connected Mode then as the token is stored in its secure storage.

This adds an extension point to be implemented by us or other third parties to the CORE bundle. It also consolidates all the Eclipse Equinox Security logic to the CORE bundle as it was implemented a bit weak leaking into the UI bundle.
Nothing changes on the current behavior, we still rely on Eclipse Equinox Security for now.

[SLE-1138]: https://sonarsource.atlassian.net/browse/SLE-1138?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vnaskos-sonar (approved):
👍

PR #862, opened 2025-03-27
SLE-803: Invalid organization should not render wizard unusabled

thahnen:
[SLE-803](https://sonarsource.atlassian.net/browse/SLE-803)

The organization is not taken into account when checking for the token anymore, which led to incorrect behavior and rendered the wizard unusable.

[SLE-803]: https://sonarsource.atlassian.net/browse/SLE-803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/org.sonarlint.eclipse.its.connected.sc/src/org/sonarlint/eclipse/its/connected/sc/SonarCloudConnectedModeTest.java
+++ b/its/org.sonarlint.eclipse.its.connected.sc/src/org/sonarlint/eclipse/its/connected/sc/SonarCloudConnectedModeTest.java
@@ -195,6 +195,19 @@ public class SonarCloudConnectedModeTest extends AbstractSonarLintTest {
 
     assertThat(organizationsPage.getOrganization()).isEqualTo(SONARCLOUD_ORGANIZATION_KEY);
 
+    // Check that when entering an invalid organization moving forward is blocked
+    var invalidOrganization = "thisOrganizationDoesNotExist";
+    organizationsPage.setOrganization(invalidOrganization);
+    assertThat(wizard.isNextEnabled()).isTrue();
+    wizard.next();
+    new WaitUntil(new DialogMessageIsExpected(wizard, "No organizations found for key: " + invalidOrganization));
+
+    wizard.back();
+    assertThat(wizard.isNextEnabled()).isTrue();
+    wizard.next();
+
+    organizationsPage = new ServerConnectionWizard.OrganizationsPage(wizard);
+    organizationsPage.waitForOrganizationsToBeFetched();
     organizationsPage.setOrganization(SONARCLOUD_ORGANIZATION_KEY);
     assertThat(wizard.isNextEnabled()).isTrue();
     wizard.next();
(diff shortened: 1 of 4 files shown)


its/org.sonarlint.eclipse.its.connected.sc/src/org/sonarlint/eclipse/its/connected/sc/SonarCloudConnectedModeTest.java
```diff
@@ -195,6 +195,19 @@ public void configureServerWithTokenAndOrganization() throws InterruptedExceptio
 
     assertThat(organizationsPage.getOrganization()).isEqualTo(SONARCLOUD_ORGANIZATION_KEY);
 
+    // Check that when entering an invalid organization moving forward is blocked
+    var invalidOrganization = "thisOrganizationDoesNotExist";
+    organizationsPage.setOrganization(invalidOrganization);
+    assertThat(wizard.isNextEnabled()).isTrue();
+    wizard.next();
+    new WaitUntil(new DialogMessageIsExpected(wizard, "No organizations found for key: " + invalidOrganization));
+
```
gabriela-trutan-sonarsource, line 204: I would add an assert here that going to the next page is not possible:

```suggestion
assertThat(wizard.isNextEnabled()).isFalse();
```

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/binding/wizard/connection/AbstractConnectionWizard.java
```diff
@@ -146,17 +145,17 @@ public final ConnectionFacade getResultServer() {
     return resultServer;
   }
 
-  protected boolean testConnection(@Nullable String organization) {
+  protected boolean testConnection(boolean doNotTestOrganization) {
```
gabriela-trutan-sonarsource, line 148: I don't think a bool makes the method more clear, on the contrary. Calling `testConnection(true)` or `testConnection(false)` does not really speak on what is the method really doing. 

Usually when a flag is introduced, it means the methods starts having 2 responsibilities.

gabriela-trutan-sonarsource (approved):
The bug was indeed fixed. But I have some feedback regarding the actual implementation.

its/org.sonarlint.eclipse.its.connected.sc/src/org/sonarlint/eclipse/its/connected/sc/SonarCloudConnectedModeTest.java
thahnen: No, as the `Next` button is not disabled. It does not react to changes in the text input fields.

gabriela-trutan-sonarsource: You are right. Not that is disabled. But that it is not possible to navigate to the next page. Would that be easy to test?

thahnen: Nope. But as clicking on the button spawns the dialog message, we already test for that.

PR #867, opened 2025-04-01
SLE-1102: Implement Sentry.io monitoring

thahnen:
[SLE-1102](https://sonarsource.atlassian.net/browse/SLE-1102)

### Summary

We introduce Sentry.io into the Eclipse plug-in by trying to capture both caught and uncaught exceptions. A `MonitoringService` is in charge of enabling and disabling Sentry.io as well as configuring it.

In order to initialize Sentry.io as early as possible, we extend the `org.eclipse.ui.startup` extension point to start with the workbench, then have everything ready for the activation of our plug-in.

The uncaught exceptions are catched by providing a status listener that listens to errors that are raised to the IDE and not caught and that contain anything relevant to our plug-in.

### Testing

We tested by provoking mimicked caught as well as uncaught exceptions from the CORE and UI bundle as well as the sub-plug-ins (e.g. Maven integration). All of it is captured correctly on Sentry!

[SLE-1102]: https://sonarsource.atlassian.net/browse/SLE-1102?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jblievremont (approved):
Nice!

PR #870, opened 2025-04-04
SLE-1157: Don't analyze when a quick fix was applied

thahnen:
[SLE-1157](https://sonarsource.atlassian.net/browse/SLE-1157)

This does not make sense due to the fact that when applying a quick fix the file is not saved and therefore the analysis runs on the pre-quick-fix-applied file.

This is a useless analysis. And in anticipation of the upcoming changes to the analysis triggering API of SonarLint Core this is removed to have less confusion.

[SLE-1157]: https://sonarsource.atlassian.net/browse/SLE-1157?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


vnaskos-sonar (approved):
👍

PR #872, opened 2025-04-07
SLE-1159: Trigger analysis of opened files on analysis properties change

thahnen:
[SLE-1159](https://sonarsource.atlassian.net/browse/SLE-1159)

When the analysis properties are changed for either the project or globally, trigger a new (automatic) analysis of the opened files to refresh based on the user changes.

[SLE-1159]: https://sonarsource.atlassian.net/browse/SLE-1159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/its/org.sonarlint.eclipse.its.standalone/src/org/sonarlint/eclipse/its/standalone/AnalyzerPropertiesTest.java
+++ b/its/org.sonarlint.eclipse.its.standalone/src/org/sonarlint/eclipse/its/standalone/AnalyzerPropertiesTest.java
@@ -1,0 +1,84 @@
+/*
+ * SonarLint for Eclipse ITs
+ * Copyright (C) 2009-2025 SonarSource SA
+ * sonarlint@sonarsource.com
+ *
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU Lesser General Public
+ * License as published by the Free Software Foundation; either
+ * version 3 of the License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+ * Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with this program; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ */
+package org.sonarlint.eclipse.its.standalone;
+
+import org.eclipse.reddeer.eclipse.ui.dialogs.PropertyDialog;
+import org.eclipse.reddeer.eclipse.ui.perspectives.JavaPerspective;
+import org.eclipse.reddeer.workbench.ui.dialogs.WorkbenchPreferenceDialog;
+import org.junit.After;
+import org.junit.Test;
+import org.sonarlint.eclipse.its.shared.AbstractSonarLintTest;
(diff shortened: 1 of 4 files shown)


its/org.sonarlint.eclipse.its.standalone/src/org/sonarlint/eclipse/its/standalone/AnalyzerPropertiesTest.java
```diff
@@ -0,0 +1,84 @@
...
+    rootProject.select();
+    var projectPreferenceDialog = new PropertyDialog(rootProject.getName());
+    projectPreferenceDialog.open();
+
+    analyzerPropertiesPreferences = new AnalyzerPropertiesPreferences(projectPreferenceDialog);
+    projectPreferenceDialog.select(analyzerPropertiesPreferences);
+
+    analyzerPropertiesPreferences.add("sonar.java.fileByFile", "false");
+    assertThat(analyzerPropertiesPreferences.getProperties().size()).isEqualTo(1);
+    doAndWaitForSonarLintAnalysisJob(projectPreferenceDialog::ok);
```
vnaskos-sonar, line 82: Technically we don't know which analysis ran, if it was for all projects or just for the project that had a property change.

vnaskos-sonar (changes requested):

org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/properties/SonarLintExtraArgumentsPreferenceAndPropertyPage.java
```diff
@@ -402,16 +404,22 @@ private void loadProperties() {
...
+      if (!sonarProperties.equals(previousProperties)) {
+        AnalysisJobsScheduler.scheduleAnalysisOfOpenFiles((ISonarLintProject) null, TriggerType.STANDALONE_CONFIG_CHANGE);
+      }
     } else {
       var projectConfig = getProjectConfig();
       if (projectConfig != null) {
         projectConfig.getExtraProperties().clear();
         projectConfig.getExtraProperties().addAll(sonarProperties);
         SonarLintCorePlugin.saveConfig(getProject(), projectConfig);
+        AnalysisJobsScheduler.scheduleAnalysisOfOpenFiles(getProject(), TriggerType.STANDALONE_CONFIG_CHANGE);
```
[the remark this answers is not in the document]
thahnen: I will nest the if inside the if, but yes I'll do so.

its/org.sonarlint.eclipse.its.standalone/src/org/sonarlint/eclipse/its/standalone/AnalyzerPropertiesTest.java
thahnen: This property was just an exemplary one I found that specifically does not change anything since we only analyze the one file here. I could have used of course a random property here as well like `internal.sonarlint.test=true`. But generally speaking the analyzer property here is out of the scope (as we don't control it) and it should only test that changing anything in the preferences, both global and local, should trigger a analysis on change.

The UI tests cannot test which analysis ran here, we only expect that when the `ok` button is clicked, that a analysis job is triggered and running. All of the ITs only have one project opened at the time or in very rare cases 2.
We also cannot check when having files from more than one project opened, which analysis ran as we send off one analysis job only and SLCORE internally split them by projects.
But this can be done quite easily manually, that is also what I did.

vnaskos-sonar (approved):
👍

PR #903, opened 2025-06-17
SLE-1180 Rework SQC ITs to eliminate dependency on existing projects

nquinquenel:


target-platforms/commons-its.target
```diff
@@ -137,36 +133,44 @@ Import-Package:        *;resolution:=optional
...
+				  <version>3.0.0</version>
+				  <type>jar</type>
+			  </dependency>
+		  </dependencies>
+	  </location>
+	  <location includeDependencyDepth="none" includeDependencyScopes="compile" includeSource="true" missingManifest="generate" type="Maven">
+		  <dependencies>
+			  <dependency>
+				  <groupId>org.apache.commons</groupId>
+				  <artifactId>commons-lang3</artifactId>
```
thahnen, line 169: You add a dependency just for `SystemUtils.IS_OS_WINDOWS`? If so, then just rely on `os.arch` instead as we only run the ITs on the CI on Linux 😕
Otherwise, this should be "okeyish" if it includes the stuff to trigger a process from the command line as well..

PR #905, opened 2025-06-18
SLE-1212 Reddeer library should correctly work on Eclipse 2025-06+

nquinquenel:
[SLE-1212](https://sonarsource.atlassian.net/browse/SLE-1212)

Starting from Eclipse 2025-06, tests are failing because Reddeer is looking for the menu “Preferences” while it’s now named “Preferences…”.

As a first workaround, we can create a new class with the fixed regex. We should take care of falling back to the previous label because this component is used across all the tests.

In the future, we should probably patch our Reddeer fork.

[SLE-1212]: https://sonarsource.atlassian.net/browse/SLE-1212?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


damien-urruty-sonarsource (approved):
Great job identifying and fixing the problem! I think the long term solution would be to fix the problem directly in Reddeer, but for now it's fine

PR #936, opened 2025-09-21
Workaround for duplicate markers on the same resource

thahnen:
This seems to be a rare corner case for when switching a branch and the new branch has the same issue on the file, a duplicate marker is found. In case there is a duplicate, we take into account the "new" one and proceed as in the following logic for all existing markers there is no new one created.

This is related to [this community forum thread](https://community.sonarsource.com/t/received-an-internal-error-occurred-during-update-issues-markers-for-project/143459).
When the internal method is invoked [here](https://github.com/SonarSource/sonarlint-eclipse/blob/11.16.0.83074/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java#L143) this fix should be sufficient as if there is a duplicate - meaning there is already a marker present - we don't create a new one [here](https://github.com/SonarSource/sonarlint-eclipse/blob/11.16.0.83074/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java#L227-L230).
In case a two markers are available for the exact same issue id, the old one is deleted (possibly a leftover).

Writing an integration test for this would be a big effort IMO tho, that's why I choose not to. Basically have a Git repo with two branches, one carrying over an issue from the first branch, then analyzing both branches in order (basically re-doing the commit CIs) and then in the IDE bind the project and switch branches.

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/SonarLintMarkerUpdater.java
@@ -129,8 +129,8 @@ public class SonarLintMarkerUpdater {
        *  which have the same issue on the same resource/file and optionally different lines.
        *  In case there is a duplicated marker, we use the new one for merging them together. This does not resolve the
        *  root issue tho but is a valid workaround.
-       *  As later in the internally invoked method "createOrUpdateMarkers(...)" for existing markers no new one is
-       *  created, we don't have to delete the duplicate here.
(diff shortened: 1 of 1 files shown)


damien-urruty-sonarsource:
hey @thahnen! I think this problem is in fact related to [SLCORE-1589](https://sonarsource.atlassian.net/browse/SLCORE-1589). There was a bug with issue matching, leading to publish 2 different issues with the same ID, in rare cases. This should be fixed in the next release. We will monitor the effect, and hopefully we can close this PR. Thanks for having a look :mage: !

PR #1079, opened 2026-07-07
SLE-1550 Fix SonarQube issues

damien-urruty-sonarsource:
----
## Summary by Gitar

- **Code quality improvements:**
  - Added private constructor to `JobHelpers` to prevent instantiation of utility class.
  - Handled `InterruptedException` by restoring the interrupted status in `waitForJobsToComplete` and polling loops.
- **SonarQube issue resolution:**
  - Replaced inefficient regex usage with `String.contains()` in `LaunchJobMatcher` and workspace job predicates.

<sub>This will update automatically on new commits.</sub>

### DIFF
--- a/org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/tests/common/JobHelpers.java
+++ b/org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/tests/common/JobHelpers.java
@@ -41,6 +41,8 @@ public class JobHelpers {
   public static void waitForJobsToComplete() {
     try {
       waitForJobsToComplete(new NullProgressMonitor());
+    } catch (InterruptedException e) {
+      Thread.currentThread().interrupt();
     } catch (Exception ex) {
       throw new IllegalStateException(ex);
     }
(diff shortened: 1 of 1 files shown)


org.sonarlint.eclipse.core.tests/src/test/java/org/sonarlint/eclipse/tests/common/JobHelpers.java
```diff
@@ -103,6 +106,7 @@ public static void waitForJobs(Predicate<Job> matcher, int maxWaitMillis) {
         Thread.sleep(POLLING_DELAY);
       } catch (InterruptedException e) {
         // ignore and keep waiting
+        Thread.currentThread().interrupt();
```
Krosovok, line 108: I think gitar might be right. If we got an interrupt here it means that current thread was interrupted for some reason.

damien-urruty-sonarsource: I missed the comment from Gitar, usually it creates inline comments, not sure why this time it was only in the summary. I fixed it as per the recommendation

Krosovok (approved):

PR #1083, opened 2026-07-08
SLE-1553 Batch issue marker updates into a single Eclipse job

nquinquenel:
----
## Summary by Gitar

- **Refactored marker update process:**
  - Converted `IssuesMarkerUpdateJob` into a singleton to queue and batch multiple project marker updates into a single job.
  - Updated `SonarLintEclipseHeadlessRpcClient` to add new marker update requests to the `IssuesMarkerUpdateJob` queue instead of scheduling individual jobs.

<sub>This will update automatically on new commits.</sub>

### DIFF
--- a/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/IssuesMarkerUpdateJob.java
+++ b/org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/IssuesMarkerUpdateJob.java
@@ -20,9 +20,13 @@
 package org.sonarlint.eclipse.core.internal.jobs;
 
 import java.net.URI;
+import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicReference;
 import org.eclipse.core.resources.ResourcesPlugin;
 import org.eclipse.core.runtime.CoreException;
 import org.eclipse.core.runtime.IProgressMonitor;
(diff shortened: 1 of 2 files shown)


damien-urruty-sonarsource (approved):

org.sonarlint.eclipse.core/src/org/sonarlint/eclipse/core/internal/jobs/IssuesMarkerUpdateJob.java
```diff
@@ -36,46 +40,97 @@
...
+
+  private final Queue<Request> workQueue = new ConcurrentLinkedQueue<>();
+
+  private IssuesMarkerUpdateJob() {
+    super("Update issues markers for projects");
+  }
+
+  public void add(ISonarLintProject project, Map<URI, List<RaisedIssueDto>> issuesByFileUri, boolean issuesAreOnTheFly) {
+    workQueue.add(new Request(project, issuesByFileUri, issuesAreOnTheFly));
+    schedule();
```
fedejeanne, line 54: One thing that could happen if one schedules without a `delay` is that rapid bursts of requests may still be processed separately i.e. the job could be scheduled, run and put to wait for each request. The delay (sometimes also called a "debounce" ) increases the probability of more requests reaching the working queue before the job even runs.

It is difficult to find the sweet spot but I suggest you try with something that is really short for the user to even notice but long enough to catch at least _some_ requests in the time window. Say `100ms`?

You don't have to do it [the way I proposed](https://community.sonarsource.com/t/update-issues-markers-for-project-spawns-lots-of-jobs-and-delays-execution-of-other-tasks/183578/5?u=fedejeanne) ...

```java
// By setting a lower priority, chances are that several requests are going
// to be processed together since they may not come in batches but one after
// the other in a short period of time.
setPriority(DECORATE);
```

but you may simply use `schedule(100)` instead of `schedule()`.

fedejeanne (commented):
Thank you for the fix!

I will try to test this and will report back in [my original post](https://community.sonarsource.com/t/update-issues-markers-for-project-spawns-lots-of-jobs-and-delays-execution-of-other-tasks/183578) in case I find something.

I only have one small suggestion regarding using a delay to reduce the probability of rapid request bursts from scheduling the job several times.

PR #1098, opened 2026-08-04
Added asyncExec guard for autobuild disabled popup

r-mennig:
### Problem description

When Autobuilding is disabled, Sonar shows a popup informing the user that results may be inaccurate. However, if multiple build commands are issued at the same time, it occasionally happens that this popup is shown multiple times (sadly we could not find a reliable reproduction for this).

image

Looking into the code shows that the cause of this problem happens in `NoAutomaticBuildWarningPopup.displayPopupIfNotIgnored()`:

image

Checking `PopupUtils.popupCurrentlyDisplayed(Class<? extends AbstractNotificationPopup>)` should guard against this popup being shown multiple times. However, this method is not checked again by the time the asynchronously scheduled runnable is actually executed, which leads to the described problem (This can also be reproduced by calling `NoAutomaticBuildWarningPopup.displayPopupIfNotIgnored()` in a loop repeatedly).

### Proposed solution

This PR combines the `isAutoBuilding()`, `popupCurrentlyDisplayed(...)` and `noAutomaticBuildWarning()` checks into one method which is re-checked inside the async runnable, guarding against the popup being shown multiple times.

### Further information

> Provide a unit test for any code you changed

I could not find a unit test bundle for this popup.

### DIFF
--- a/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/popup/NoAutomaticBuildWarningPopup.java
+++ b/org.sonarlint.eclipse.ui/src/org/sonarlint/eclipse/ui/internal/popup/NoAutomaticBuildWarningPopup.java
@@ -70,13 +70,15 @@ public class NoAutomaticBuildWarningPopup extends AbstractSonarLintPopup {
 
   /** This way everyone calling the pop-up does not have to handle it being actually displayed or not */
   public static void displayPopupIfNotIgnored() {
-    if (ResourcesPlugin.getWorkspace().getDescription().isAutoBuilding()
-      || PopupUtils.popupCurrentlyDisplayed(NoAutomaticBuildWarningPopup.class)
-      || SonarLintGlobalConfiguration.noAutomaticBuildWarning() ) {
+    if (!shouldShowPopup()) {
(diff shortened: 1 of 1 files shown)


nquinquenel:
Thanks a lot for the clear write-up and the proposed fix, I think the diagnosis is correct.

A couple of notes on process and scope:

Our contribution guidelines have become more restrictive. We no longer encourage external pull requests for this repository; we prefer to take the report and implement the change ourselves so it matches our roadmap, testing, and review process, as well as our legal policies.

On the fix itself: this PR only guards `NoAutomaticBuildWarningPopup`. The same race exists for the other async notification popups (`InvalidTokenPopup`, `LanguageFromConnectedModePopup`, `NewerVersionAvailablePopup`, `NoBindingSuggestionFoundPopup`, `ReleaseNotesPopup`). We opened 
[#1108](https://github.com/SonarSource/sonarlint-eclipse/pull/1108), which applies a shared `PopupUtils.scheduleAsyncDisplay(...)` helper to all of those and adds unit tests for the race.

We’ll handle the rest from there. Thanks again for reporting it and for the initial patch!

PR #1115, opened 2026-09-17
SLE-1622 Remove Kotlin OSGi bundle after dropping Xodus

damien-urruty-sonarsource:
## Summary
- Update sonarlint-core to 11.11.0.87260, the build that dropped Xodus (`SLCORE-2584`)
- Remove the Kotlin OSGi bundle from the ITS target platform, which was only needed because of that dependency


damien-urruty-sonarsource:
Closing: Kotlin still cannot be removed because connected ITs need it for Orchestrator → OkHttp → Okio.

